PR #108, opened 2015-02-07
Use the correct mode when generating anchors.

Zegnat:
Apparently `render_name` doesn’t inherit the value of `mode` from the `render` macro, so we have to specifically pass it along. This will fix anchors for the games that exist in both modes, previously they wouldn’t include the mode and always link to the clones and never to the ‘inspired by’ section.


piranha:
Oh, that's a nice finding, thank you! :)

PR #109, opened 2015-02-10
added clone pac-man

nonothing:


piranha:
There are hundreds of pac-man clones. Sorry, not really interested.

PR #125, opened 2015-03-30
Fixed Wikipedia links

wonea:
Fixed many Wikipedia links either broken, or linking to something not todo with a video game


piranha:
Thanks!

PR #142, opened 2015-04-07
Added Populous, R-Type & StarShatter clones

wonea:


piranha:
That's an endless stream of updates! Thanks. :-)

Issue #144, opened 2015-04-07
Entombed 1982 missing Wikipedia page


wonea:
http://www.newbreedsoftware.com/entombed/

Will add once the page exists

wonea:
Added Wikipedia page https://en.wikipedia.org/wiki/Entombed_%281982_video_game%29

Issue #137, opened 2015-04-05
EarthShaker


wonea:
Cone but no Wikipedia page

https://github.com/ant512/EarthShakerDS

wonea:
Added Wikpedia page

PR #146, opened 2015-04-08
Added various games

### LINKED ISSUE #136: Survivor 1982 Commodore 64 video game remake
Currently no Wikipedia article for 1982 version, therefore can't add

https://github.com/scottschiller/SURVIVOR

2015-04-05 14_16_09-survivor - wikipedia the free encyclopedia

wonea:
Fixes #136 & #137 & #144

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -1530,6 +1530,13 @@
       media:
         - url: http://www.hackcraft.de/games/linwarrior_3d/screenshots/20110417.jpg
 
+- name: MechCommander 2
+  clones:
+    - name: MechCommander 2 Omnitech
+      added: 2015-04-08
+      repo: https://github.com/Echelon9/mechcommander2-open
+      info: halted development, C++, mixed licence; Shared Source Limited Permissive License & GPL
+
 - name: Mega Lo Mania
   clones:
     - name: Gigalomania
--- a/games.yaml
+++ b/games.yaml
@@ -2086,6 +2086,15 @@
       media:
         - image: http://a.fsdn.com/con/app/proj/methane/screenshots/14915.jpg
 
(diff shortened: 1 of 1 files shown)


games.yaml
```diff
@@ -2786,6 +2855,14 @@
           image: http://www.oolite.org/images/gallery/coriolis!jpg
         - url: http://www.oolite.org/images/gallery/full/advanced-nav-array
           image: http://www.oolite.org/images/gallery/advanced-nav-array
+  clones:
+    - name: Mission ELITE: The Battle For Birera
```
piranha: You have to put quote marks around string if it contains a colon, so should be `- name: "Mission ELITE: The Battle For Birera"`

wonea:
I think its not building because some nodes have both; clones & reimplementations

games.yaml
```diff
@@ -1839,6 +1940,21 @@
         - image: http://freesiege.sourceforge.net/screens/screen0.png
         - image: http://freesiege.sourceforge.net/screens/screen1.png
 
+- name: Siege of Avalon
+  clones:
+    name: Siege of Avalon Open Source
```
piranha: You forgot to prepend name with `-` here. :) Unfortunately YAML is a bit brittle. I tried to improve error reporting a bit.

piranha:
Now Travis report is much better.

games.yaml
wonea: Thanks for your help. Start going cross eyed after a while!

piranha:
Cool!

I wonder if I need to rework this in a static typed language just so that error reporting is better. :))

wonea:
More reporting on build errors is always helpful. Lines number usually are enough, but sometimes the obvious is not obvious!

Issue #150, opened 2015-04-10
Transplant


piranha:
a open source game that is inspired by a game called "Transplant" from the Commodore Amiga computer.
Please check it out: https://spaceshipduel.codeplex.com/

wonea:
Will add but I'm fighting a couple of page deletions at the moment on Wikipedia. Wikipedians don't like short pages on obscure games :-(

piranha:
Well, I guess we could live without pages on Wikipedia. :) Especially if you know a page describing the game, putting `[Name, "http://some/link"]` will point to that place.

PR #151, opened 2015-04-10
More games added

### LINKED ISSUE #149: Ace of Spades
OpenSpades ( https://sites.google.com/a/yvt.jp/openspades/downloads ), an open alternative to Ace of Spades

wonea:
Fixes #149 & #150

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -243,6 +243,18 @@
         - image: https://a.fsdn.com/con/app/proj/openastromenace/screenshots/am7.jpg
           url: https://a.fsdn.com/con/app/proj/openastromenace/screenshots/am7.jpg/182/137
 
+- name: [Babaliba, "Dinamic Software"]
+  clones:
+    - name: Babaliba
+      added: 2015-04-13
+      url: http://compiler.speccy.org/index.php?lang=eng&game=babaliba
+      repo: http://compiler.speccy.org/files/sources/Babaliba_Source.zip
+      info: Microsoft Visual FoxPro, no specified licence
+      media:
+        - image: http://compiler.speccy.org/images/screens/babaliba-scr1.png
+        - image: http://compiler.speccy.org/images/screens/babaliba-scr2.png
+        - image: http://compiler.speccy.org/images/screens/babaliba-scr3.png
+
 - names:
     - "Baldur's Gate series"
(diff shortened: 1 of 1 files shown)


wonea:
Okay, I'm done for now.

wonea:
Okay, better finish now.

piranha:
Canabalt-iOS is not a clone, it's an exact source of the game, isn't it?

wonea:
Originally, there has been a lot of activity since it was open sourced. Noticed a pull request to update it for XCode 4

piranha:
Yeah, but still, it's not a clone... :)

AGNUcius:
Hey all, I've been archiving links to Open Source video games for a while
now.

I am considering using the data from OSGameClones to make a more complete
list.

Just wondering if there are any plans to expand the scope of this project
to all Open Source games.

Thanks,
Patrick Anderson
My current list: http://crosscrowdpredictiveproduction.github.io/game.htm

wonea:
Wow, that's an impressive list. Love the commentary.

piranha:
Making an informational project about all open source games is certainly a good idea, but then it would be much bigger than OSGC, requiring much more time to maintain and curate the content - especially given that quite a lot of the games wouldn't have external sources (like Wikipedia) to describe them well, and having descriptions would help.

So on one hand I like the idea, on the other - I'm afraid I'm too short on time to make a viable take on that.

I'm wondering how could we handle that, if we decide to go that road? Single YAML as a database would be really hard to manage. I guess we could split in a few files, but how would we divide them? One alternative coming to mind is standard rdbms-backed site, but that would be an overkill and much more painful to maintain (OSGC is literally 0 upkeep right now).

And then the interface. Any thoughts about the interface?

If we could decide on those points, I guess we could convert the project into 'open source games catalogue' (haha, still osgc :)), and then osgameclones.com would just display a subset of games, or something like that. :)

wonea:
I think we should really make sure we reference Wikipedia. It's a good bar to cross, I don't mind creating the pages as required (have 6 pages needing creation at the moment).

'open source games catalogue' is a good idea. Perhaps we have another node/leaf in the yaml to sit alongside clones / reimplemention - original. Keep them in a separate yaml file.

PR #161, opened 2015-04-15
Added GoldenEye 007 clone

wonea:


wonea:
Nice find, eh?

piranha:
Yeah, that one's good. :-)

PR #162, opened 2015-04-15
Added Pirates! clone

wonea:


games.yaml
```diff
@@ -2496,6 +2558,35 @@
       media:
         - image: http://freecivac.sourceforge.net/pics/govchange.png
 
+- name: ["Sid Meier's Colonization", Sid Meier's Colonization]
```
piranha: I wonder why did it build in the first place, considering there are no quotes around link to wiki. :)

PR #164, opened 2015-04-16
Mostly old remakes

wonea:


games.yaml
```diff
@@ -2558,7 +2687,7 @@
       media:
         - image: http://freecivac.sourceforge.net/pics/govchange.png
 
-- name: ["Sid Meier's Colonization", Sid Meier's Colonization]
+- name: ["Sid Meier's Colonization", "Sid Meier's Colonization"]
```
piranha: I think we can get away with `name: "Sid Meier's Colonization'` here.

PR #168, opened 2015-05-02
New assortment of games

wonea:


piranha:
I really have no words. :) You're on fire, man!

PR #177, opened 2015-05-26
Update fall down reimplementations

cxong:


wonea:
@piranha are we okay with the game lookup going a different website other than Wikipedia or MobyGames?

http://web.eecs.umich.edu/~gameprof/gamewiki/index.php/Fall_Down

piranha:
I was thinking about that and... yeah, I don't know, should we strive for linking to clones of the games which people are going to search for? Or just anything that ever existed? That was a hard question for me - I even resisted adding Minecraft clones at first since "it wasn't dead yet by itself".

cxong:
Fall down doesn't have a wikipedia page but it is notable; it and its sequel were number 1 app store games at one point. Perhaps we should link to the app store instead?

wonea:
@piranha personally I think we should list everything which is a clone or reimplementation. How do we quantify what people are interested in? I love old rare Amiga games, but that's just me! ;-) Get a little exited when I see a rare public domain game reimplementation! hehe

@cxong nice work on the game! Congratulations! I would only prefer MobyGames to be referenced as it is a possible seen as more definitive than a university web site. Other than that I personally welcome the addition.

piranha:
Ok, makes sense. :-)

PR #185, opened 2015-06-01
Added Movie Business clone

wonea:
Couldn't find a Wikipedia or MobyGames page, and contacted the author for help

https://github.com/teruk/movbizz/issues/4#issuecomment-107454052

So linking to Gamebase64

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -2711,6 +2711,14 @@
       media:
         - image: http://a.pomf.se/jksvcq.png
 
+- name: [Movie Business, "http://www.gamebase64.com/game.php?id=14674&d=18"]
+  clones:
+    - name: movbizz
+      repo: https://github.com/teruk/movbizz
+      added: 2015-06-01
+      info: active development, (PHP & JavaScript), MIT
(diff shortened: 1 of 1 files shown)


piranha:
Haha, I've even played and it seems when you advertise through a TV, it can't finish turn. :-)

teruk:
really? can you provide me with more information? I can't recreate this error

piranha:
Well, I created a movie, and then waited for a round, made 1 TV advertising, skipped two rounds and it showed me an error page with text 'Whoops, looks like something went wrong.'

teruk:
It should be fixed now. It had actually nothing to do with the advertisement.

piranha:
Heh, I see. :) Though to be honest I can't understand how am I supposed to make money. All I manage to do is spend all my money on a movie and then it never makes any decent profit. :)

wonea:
Are you trying to make Waterworld? http://en.wikipedia.org/wiki/Waterworld

teruk:
I know the problem, the balancing is still something I have to work on. But you would have the same problem in the original game ;-)

PR #187, opened 2015-06-02
Couple of Construct2 games

wonea:
Asked about source code here

https://github.com/jvlppm/pucpr-ld-parodius/issues/1#issuecomment-107773628


piranha:
I wonder if we should render `info` through markdown and make `Construct2` a link, because I don't think that's universally understandable. Any thoughts?

piranha:
Or maybe just render it as HTML. So that tags can be used there.

wonea:
Yes, that sounds like a good idea. Love3D could be a similar problem.

piranha:
Turned out it already works so :-)

PR #192, opened 2015-06-04
Update the repo and url entry of projects that left sourceforge.

vloup:
I just had a look at the projects that moved out of sourceforge as you may have heard recent news. I saw that Maxit, OpenTomb and OpenDungeons moved to GitHub. Here is the patch with updated url, repo and images.


piranha:
Heh, sourceforge has really failed, and it starts to show. Thanks! :)

PR #204, opened 2015-07-02
Add Quake 1 clones

vloup:
This add the most 3 active Quake projects: DarkPlaces, ezQuake and QuakeSpasm. I also re-arranged the whole list to be alphabetically ordered.
I am open for any correction or questions.
Have a good day.


piranha:
Thanks!

PR #219, opened 2015-07-22
Add xash3d, a Half-Life clone

szaromrok:


piranha:
That's not open source, that's a stolen Valve code.

PR #221, opened 2015-07-29
Add Doom 3 clones.

vloup:
This is pretty self explanatory.
The first port is only for the first edition only and the second is only for the BFG edition.

I am open to change anything that will suit your needs.
Have a good day.


piranha:
Cool, thanks!

PR #222, opened 2015-08-07
[fix] Removed legacy project glPortal2. Fixed broken screenshots.

hhirsch:


piranha:
Thanks!

hhirsch:
Pleasure. Thanks for all the impressions.

Issue #223, opened 2015-08-08
Lode Runner remake version (HTML)


SimonHung:
I porting this game from a old book with lode runner source code in C 
GitHub: https://github.com/SimonHung/LodeRunner_TotalRecall
Web Page: http://LodeRunnerWebGame.com
Game URL:  http://LodeRunnerWebGame.com/game

Simon Hung

PR #225, opened 2015-08-15
Added Lode Runner clone

wonea:


wonea:
Fixes #223

piranha:
Life became so good since I made "deploy directly from travis" work, daaamn. :-))

PR #226, opened 2015-08-18
Updated X-Com games

SupSuper:
- Corrected infos
- Added X-COM: Terror from the Deep to OpenXcom and UFO2000
- Separated OpenApoc into its own X-COM: Apocalypse entry


piranha:
Thanks!

Issue #234, opened 2015-09-26
Dungeon Master Port and Clones


Ty-Guy5:
Neat list we have here!

There's a C++ Port for the old Atari game, Dungeon Master, and its sequel, Chaos Strikes Back. It's called CSBwin, and it's code is (technically) open source. Following the links on the main page at http://dungeon-master.com/ will give you CBSwin (high fidelity to original) and a few other clones (RTC, Return to Chaos, is the big one). It's been modified to allow creation of custom dungeons, some of which are very fun! I even wrote a dungeon myself, several years back.

I would recommend linking to http://dungeon-master.com/ for an intro, links to the clones, and the community forums. 
Clicking "CSBwin" from there brings you to http://dmweb.free.fr/?q=node/851,  with the list of the downloadable (playable) files, including versions for Windows, Linux, Mac, and PocketPC. 
Also link to http://www.ragingmole.com/RTC/ for RTC.

wonea:
Thanks for the heads up about this game. Always great to find clones for games we haven't listed here before!.

Issue #191, opened 2015-06-03
3 excellent opensource zelda games


frostworx:
maybe interesting:
http://www.zeldaroth.fr/us/
gentoo ebuilds are in gamerlay overlay btw

wonea:
Thanks for the links, sorry it's taken a while but I've finally added the links.

PR #235, opened 2015-09-27
Added Dungeon Master clone

wonea:
Note bizarre licence #234 & #191

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -1295,7 +1295,6 @@
       added: 2014-09-05
       info: development halted, C++, MIT
 
-
 - name: Dungeon Keeper
   clones:
     - name: KeeperFX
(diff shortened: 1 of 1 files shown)


piranha:
Should we not care about that? There is xash3d, which is basically stolen HL sources, and I'm not going to add that, should we maybe do the same with DM?

wonea:
It's not stolen, but reverse engineered. Excerpt from the project homepage:

"Finally, after many years, I got my hands on the binary executable for the game. I wrote a disassembler to turn it into human-readable op-codes and proceeded to translate it to C++ using the Microsoft Version 6.0 C++ compiler. Eight hours a day for six months. About 120,000 lines of pseudo-assembly language. Crazy thing to do. But it works."

Read the COPYING file in the repo file, details the original author being a little hostile to the project;

This game is not GPL.
It's not even Freeware.
It's a port.


From: Paul R. Stevens <prsteven@facstaff.wisc.edu>                                                                                                         
To: Erik Svanberg <svanberg@acc.umu.se>                                                                                                                    
Subject: CSB: COPYING                                                                                                                                      
                                                                                                                                                           
This game is really not Freeware.  It is Stolenware.                                                                                                       
Here is what Wayne Holder wrote about CSBwin:                                                                                                              
                                                                                                                                                           
>Please be aware that DM and CSB are both copyrighted and trademarked                                                                                      
>products that belong to Software Heaven, Inc (SHI).  The FTL swoosh is                                                                                    
>also a trademark of SHI as are the images in the program, the design of                                                                                   
>the game data structures and numerous other aspects of the original                                                                                       
>product.  You and your friend have no right to convert this intellectual                                                                                  
>property to GPL because it belongs to SHI.                                                                                                                
>                                                                                                                                                          
>Wayne                                                                                                                                                     
                                                                                                                                                           
Might he also object to the 'Freeware' label?  I don't                                                                                                     
know the exact connotations of these words.

piranha:
Ok, that makes it somewhat more or less like TTDpatch, I guess. I'm okay with that. :)

PR #237, opened 2015-09-28
Added an interesting Minesweeper clone

wonea:
F# and iOS how can we say no?


piranha:
F# is so much yes!! :)

PR #238, opened 2015-09-29
Don't add link if game has no url

cxong:
Lots of games don't have URLs so the link created is simply the current page - osgameclones.com - so clicking them just reloads the page, which is confusing.


piranha:
Thanks!

PR #242, opened 2015-10-21
Add Xash3D SDL(Half-Life)

a1batross:

PR #243, opened 2015-10-21
Remove Sing engine(Half-Life)

a1batross:
According to #242 PR, Xash3D is not open source. (but it really open, maybe not free, but open) 
Sing engine is a fork of Xash3D. So it must be removed.

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -1907,15 +1907,6 @@
       media:
         - raw: <iframe width="420" height="315" src="//www.youtube.com/embed/JBjdzWomC4M?rel=0" frameborder="0" allowfullscreen></iframe>
 
-
-- name: [Half-Life, Half-Life (video game)]
-  clones:
-    - name: Sing
-      url: http://code.google.com/p/sing-engine/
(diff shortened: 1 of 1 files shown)


piranha:
Damn, really? Thanks!

As for open/not open: https://en.wikipedia.org/wiki/Open_source

... open source as a development model promotes universal access via a free license to a product's design or blueprint, and universal redistribution of that design or blueprint...

And xash3d is just available source, but it can't be distributed under free license because it's stolen proprietary product.

a1batross:
Yes. Half of a code is under GPLv3, other is under HLSDK license which is not free. (but already marked as GPLv3, thanks, Uncle Mike!)

piranha:
You can't just add some code to leaked SDK and declare it GPLv3. It doesn't work like that.

a1batross:
There is no code from _leaked_ SDK. Only code that Valve released as HLSDK.

piranha:
Ok, that's a little bit better, but you still can't do that. Neither GPL nor HL SDK license permit this.

a1batross:
I know. *cries*

PR #244, opened 2015-11-23
Updated OpenRA website and screenshots

Mailaender:
Website changed from http://www.open-ra.org to http://www.openra.net a while ago.

I also updated the screenshots to reflect the progress of the game engine and highlight all the supported games/mods as well as the modernization regarding UI compared to the originals:

in-game map editor Red Alert mod
Tiberian Dawn mod UI and support power A10 strike
Replay system and range indicators in the Dune 2000 mod

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -848,13 +848,13 @@
     - Dune 2000
   clones:
     - name: OpenRA
-      url: http://www.open-ra.org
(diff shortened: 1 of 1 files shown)


piranha:
Nice, thanks!

PR #245, opened 2015-12-06
Update information about VCMI

ArseniyShestakov:
* The repository has moved to GitHub
* Can be considered as playable now (though still more work to do on multiplayer, AI)


piranha:
Argh, travis changed something, so it won't build without some changes.

Thanks for the update though!

PR #249, opened 2016-01-01
Added The Battle for Wesnoth

absorber:
My first PR of the year. Happy 2016 :)

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -2882,7 +2882,7 @@
     - name: The Battle for Wesnoth
       url: https://www.wesnoth.org/
       repo: http://wiki.wesnoth.org/WesnothRepository
-      info: The Battle for Wesnoth is a Free, turn-based tactical strategy game with a high fantasy theme, featuring both single-player, and online/hotseat multiplayer combat.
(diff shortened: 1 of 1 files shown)


wonea:
Thank you. Battle for Wesnoth cannot be a clone of itself. According to Wikipedia's article it could be seen as an reimplementation of Masters of Monsters & Warsong.

https://en.wikipedia.org/wiki/The_Battle_for_Wesnoth

absorber:
@wonea I've checked out the gameplay of [Masters of Monsters](https://www.youtube.com/watch?v=Ky0hrCdzrUM) and it does look very similar. The only difference I've noticed is that the "battle mode" is not included in The Battle for Wesnoth and is done all in the "map" interface.

I've done a new commit on this branch, is it better now?

wonea:
Looking better, one more more correction. Info should contain code details.

info: active development, C++ & Lua, GPL2

absorber:
@wonea I've updated The Battle For Wesnoth info and also made a new [PR](https://github.com/piranha/osgameclones/pull/253) in which the `info` field example from `README.md` so that new contributors can add relevant info from the start.

piranha:
Cool, thanks! That's quite a list though. :-)

PR #258, opened 2016-02-05
Fixed Ancient Beast screenies

DreadKnight:
added more permanent links


DreadKnight:
Awful in practice it seems. Will make another pull request fixing stuff.

PR #260, opened 2016-02-07
Update games.yaml

5tefan:
Updating info about BZFlag, repo is now on Github


piranha:
Cool! But please fix the error:

`Error: ['Battle Zone', 'Battlezone_(1980_video_game)', 'Battle Zone Capture The Flag'] should be a string or a list of two strings`.

It's a name of a game plus a name of page on Wikipedia, so there is no place for the third item there.

5tefan:
Ah, fixed!

piranha:
Thanks!

Issue #256, opened 2016-01-19
PARPG - Fallout Inspired


DomtronVox:
Fallout 1 and 2 __inspired__. No longer actively developed. Written in python using an FOSS C++ framework (FIFE www.fifengine.de). I believe the code license is GPL/BY-SA [forum discussion link](http://forums.parpg.net/index.php?topic=770.0)

Here is the web site link http://blog.parpg.net/

piranha:
Done in #261

Issue #255, opened 2016-01-15
FOnline engine (Fallout 1/2)


Anthony-Gaudino:
FOnline engine was created by a Russian group of Fallout fans, this engine can use assets from Fallout 1/2 and other games.

Many improvements were made in relation to the original Fallout 1/2 engine, including the use of 3D models and multiplayer.

This engine is used on many Fallout style community games, including multiplayer and single player games, but most of them didn't changed the engine core much. Most games are "clones" of Fallout Online and there's a "clone" of Fallout Van Buren, both games were never released by Interplay.

There are many different games using the engine:
* FOnline 2
* FOnline 2238
* FOnline Reloaded
* FOnline: Ashes of Phoenix
* Van Buren (http://www.moddb.com/games/f3-van-buren     single player)
* and many others...

There are many source code repositories of the engine, there's some on GitHub:
* Source of FOnline 2238 and tools - https://github.com/rotators
* https://github.com/alexknvl/fonline

wonea:
Thanks for information, this is a really great find. Thank you!

piranha:
Done in #261

PR #261, opened 2016-02-08
A few new games and fixes

### LINKED ISSUE #250: TrueCraft does not use a custom license
It uses the MIT license.

https://github.com/SirCmpwn/TrueCraft/blob/master/LICENSE

wonea:
Fixes #250 #256 #255

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -2639,7 +2639,7 @@
         - image: http://perso.b2b2c.ca/sarrazip/dev/images/afternoonstalker-1.png
           url: http://perso.b2b2c.ca/sarrazip/dev/images/afternoonstalker-1-tn.png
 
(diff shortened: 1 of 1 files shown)


games.yaml
```diff
@@ -2616,6 +2631,16 @@
         - image: http://perso.b2b2c.ca/sarrazip/dev/images/afternoonstalker-1.png
           url: http://perso.b2b2c.ca/sarrazip/dev/images/afternoonstalker-1-tn.png
 
+- name: [Octopus, https://en.wikipedia.org/wiki/List_of_Game_%26_Watch_games#Octopus]
```
piranha: I think it can be written without `https://en.wikipedia.org/wiki/` prefix

PR #262, opened 2016-02-09
Add link to the list of games made using ASCII graphics.

ligurio:
My list of opensource games is similar to your list, but I like text-based games :) Please add link to my list to your site.


piranha:
Cool! :-)

PR #264, opened 2016-02-10
Update games.yaml

pixe1f10w:


piranha:
Xash3D is not open source, it was developed from the sources of actual Half-Life and it's status is somewhat controversial.

pixe1f10w:
Sure, I've already thought about that fact.

PR #265, opened 2016-02-16
Update games.yaml

pgoca:
All of the listed projects inspired by SimCity are unfortunately abandoned.


piranha:
Eh... that's too bad, thanks for the update!

PR #272, opened 2016-03-06
Added Netacka and rawgl

Mailaender:
Fixes a regression from https://github.com/piranha/osgameclones/pull/270 and adds rawgl which seems to be what New RAW has forked and netacka which is a multiplayer version of zattacka.


wonea:
Thanks!

PR #280, opened 2016-03-23
Add issue template

cxong:
Using [GitHub's issue templates](https://github.com/blog/2111-issue-and-pull-request-templates), this PR means whenever someone creates a new issue, the body is automatically filled in like this:

image

This is so that the user can provide as much information as possible, making it easier and faster to resolve their issues.


wonea:
This is a great idea, I for one vote this be merged.

piranha:
That's a really good idea, thanks!

PR #281, opened 2016-03-24
Added uMario

Fighter19:
#Added uMario, a good and as far as i can tell complete Mario Clone.


cxong:
Looks like a great clone, unfortunately it's using assets owned by Nintendo. Will this cause problems? Should this information be included in the entry?

Fighter19:
I don't really think it is, especially not for this repo, as this is just a collection of projects.
It depends on the region you're living in. However the worst that could happen is as far as i can tell, the uMario repo getting taken down by Nintendo (see Mario64 HD remake, in Unity) and this entry has to be undone.

wonea:
Thanks, I think this entry is fine. A lot of entries on here require the original assets to run.

PR #289, opened 2016-04-18
Tags support 

nikuda:
The game `info` key was being populated with similar data that would be more useful in its own keys. This PR creates the new keys to free up `info` for general comments and uses the new keys in a tagging system. The change is additive, the data formatted in the old way will still show, it just won't have the tags.

**New keys**

`status`: (string) status of the project (active, halted ...)
`license`: (string) project license
`lang`: (string|list) programming languages used in the project
`framework`: (string|list) programming frameworks and/or engines used in the project

**Screenshot**

This shows the 15 or so items which with  `info` data has been placed into the new keys. There needs to be more work into "porting' `info` from old `games.yaml` to the new format one. 

screen shot 2016-04-18 at 8 23 00 pm


nikuda:
Probably need a game `type` key too.

wonea:
I like this idea, and would happily do the necessary changes to games.yaml

I also like the idea of putting the languages first, someone might come onto this site, and think... I know C# let at glance list all the games I could work on. At the moment, yes we can search but this gives a nice visual representation.

@piranha what do you think?

nikuda:
>I also like the idea of putting the languages first, someone might come onto this site, and think... I know C# let at glance list all the games I could work on. At the moment, yes we can search but this gives a nice visual representation.

This is exactly where the idea for this came from :relaxed:  C lang in particular is difficult to filter with, even if stricter string checking was added to the search (i.e. `"C"`) it's still not as obvious and precise as tags.

If this PR is merged more work can be done on making delimited search and tags work together. Also some preliminary work is there to support related tag highlighting. This is so if you click on `C` tag it will hint at related tags so that you can further narrow down your search, i.e. Show me only projects with `C` + `Python`.

piranha:
Sorry I didn't reply sooner, was really busy in offline life! 

I like the idea very much, having structured information is always a win over having free-form one. :)

piranha:
Ok, I'm merging that, and if anybody has time and desire, we'll have to convert more games to that format :-)

PR #294, opened 2016-05-01
Added Block Attack - Rise of the Blocks

sago007:
It is added an "reimplementations". Although it is closer to the original game than for instance Crack Attack.

I did deliberately not change any of the other games in the category to make the pull request smaller.


wonea:
Thanks, great addition

PR #302, opened 2016-05-09
Sorted tag cloud, tag and general fixes

nikuda:
See commits for more info


wonea:
Let's give this a go!

PR #329, opened 2016-06-02
added Armagetron Advanced and GLtron

tsankuanglee:


wonea:
@piranha I'm not sure about this, clone of an already open sourced game?

tsankuanglee:
I guess the original version is the movie itself, so all the other games are clones of that.

piranha:
I'd say that's not exactly what we were looking for so far. The site is not for all open source games right now, as with that it would be a little bit different.

Not sure, do we need a catalogue of all open source games? Original intention of a site was even more focused that what we have right now, so maybe the way forward is to open up for all games, but... wouldn't that be a Herculean effort? :)

PR #330, opened 2016-06-02
Added NFSIISE

Mailaender:
via https://www.reddit.com/r/linux_gaming/comments/4m6683/need_for_speed_ii_se_linux_port_with_3d/

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -3323,6 +3323,19 @@
         - image: http://1amstudios.com/img/opennfs1/image13.jpg
         - image: http://1amstudios.com/img/opennfs1/image10.jpg
 
+- name: Need For Speed II SE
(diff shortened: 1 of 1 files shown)


cxong:
From reddit:

> Isn't there a legal problem here since this has code from the original binaries and is not a re-implementation?
Perhaps the intention is the swap out parts until eventually having replaced all the original parts, but legally speaking would that be considered derivative work even though at that stage it wouldn't contain any part of the original binaries?

Mailaender:
OpenTTD started out the same way. I am not a lawyer and I don't want to judge the legality, but rather add it to this collection.

wonea:
Agreed, I think the game should be in this collection.

PR #333, opened 2016-06-07
Ability to deselect tags

### LINKED ISSUE #319: How to unselect a filter?
The new filter system is nice, but how do I unselect a filter, so that every game appears again? I can only switch bewteen filters.

nikuda:
Fixes #319

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -70,12 +70,12 @@
     var game, gameTags;
 
     if (curTag === activeTag) {
-      activeTag = false;
-      document.body.className = '';
-      highlightTags(false);
+      activeTag = null;
+      removeClass(document.body, 'tags-active');
+      highlightTags(null);
     } else {
       activeTag = curTag;
-      document.body.className = 'tags-active';
+      document.body.className = (document.body.className + 'tags-active').trim();
       highlightTags(curTag);
     }
 
@@ -100,6 +100,11 @@
     var style = document.getElementById('tag-style');
     var line = '[data-name=\"' + tag + '\"] { color: #ccc; background: #444; }';
     style.innerHTML = tag ? line : '';
+  }
+
+  function removeClass(node, name) {
+    var re = new RegExp('(?:^|\\s)' + name + '(?!\\S)', 'g');
+    return node.className = node.className.replace(re, '');
   }
 
   function hasClass(node, name) {
--- a/static/main.js
+++ b/static/main.js
@@ -70,9 +70,9 @@
     var game, gameTags;
 
     if (curTag === activeTag) {
-      activeTag = null;
+      activeTag = false;
       document.body.className = '';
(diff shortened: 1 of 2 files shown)


static/main.js
```diff
@@ -57,28 +57,34 @@
...
   function onclick(e) {
     var t = e.target.hasAttribute('data-name') ? e.target : e.target.parentNode;
+    var curTag = t.getAttribute('data-name');
     var parent;
     var game, gameTags;
 
-    if (!hasClass(document.body, 'tags-active')) {
-      document.body.className += ' tags-active';
+    if (curTag === activeTag) {
+      activeTag = false;
```
piranha: I would prefer seeing `null` here. It's not a big deal, but it's semantically more correct (the same with `highlightTags` argument lower - `false` vs `null`).

static/main.js
```diff
@@ -57,28 +57,34 @@
...
+      document.body.className = '';
```
piranha: Not sure if that's good to do, since some browser extensions tend to assign classes on `document.body`. On the other hand, that's their problem, but we can do `document.body.className = document.body.className.replace(/ ?\btags-active\b/, '')` (not sure this regexp is correct, haha).

static/main.js
nikuda: Good call.

static/main.js
nikuda: Yes it's pretty destructive behavior but thought it was in a controlled env (didn't think about extensions). Adding a non-lazy solution soon.

nikuda:
Could use `classList` to add/remove class names too, but `classList` is IE10+.

cxong:
With all these extra features being added, maybe it's time to look at jquery.

nikuda:
Hmm, with the current usage as is, my vote would be to use `classList` and drop IE < 10 versus adding jQuery.

nikuda:
I've made those change to use `classList` in

nikuda:
I've made the changes to use `classList` in 01a5249c1082f8c1e5aacb21c6a055d400744ef5.

To check the previous changes without `classList` see https://github.com/piranha/osgameclones/compare/master...nikuda:82da7b5aff243583e037d7a055c3900de47b32c8

Up to you guys 😌

wonea:
I could imagine IE 10 usage for a (developers) site like this would very low

cxong:
osgameclones.com has google analytics so that should be the last word on how many people use IE <10.

My experience is that it shouldn't be ignored; quite a few people playing open source games are using older computers, using an OS like windows XP which can only use IE 8. In this case it's not a big deal because all it means is the tags don't work for them.

piranha:
So in total we have 2.19% of IE, 88% of them being IE11. I guess it's pretty save to drop support for IE<10!

PR #335, opened 2016-06-09
Removed license keys with unknown license, preventing rendering tags with no value

nikuda:
To have an easier grouping of projects with unknown license we omit the `license` key from the game data instead of specifying the key with different variations of the same thing.


wonea:
Thanks

PR #339, opened 2016-06-18
Added Herzog 3D

Mailaender:
A 3D remake of the first RTS game.


wonea:
Nice find, thank you!

PR #347, opened 2016-07-31
Added the XL Engine

Mailaender:
A potentially large, but apparently one person project. The documentation is a bit disorganized. Not sure if I got everything correctly. See also http://xlengine.com/xl-engine-readme/ which is a bit hidden in the blog.

### DIFF
--- a/games.yaml
+++ b/games.yaml
@@ -1901,14 +1901,6 @@
       license: GPL3
       status: not playable
       added: 2015-07-17
-    - name: XL-Engine
-      repo: https://github.com/luciusDXL/XL-Engine
-      url: http://xlengine.com/
-      added: 2016-04-24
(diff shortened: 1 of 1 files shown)


piranha:
There is an entry for XL engine already: http://osgameclones.com/#the-elder-scrolls-ii:-daggerfall-clones

Not sure what's the best way to deal with such situation, but leaving XL Engine in two places seems to be not the one. :) Maybe remove it from the current place then?

Mailaender:
Updated. This is indeed a weird combination of games in this engine. I was looking for a way to play Outlaws and Dark Forces so I found https://darkxl.wordpress.com/ which pointed me there.

piranha:
Thanks!

PR #361, opened 2016-10-08
Add dune2themaker4j

stefanhendriks:


wonea:
Thanks, I've updated the current entry.

stefanhendriks:
Thx @wonea - I checked the `master` branch but I did not see any changes. Did you forget to merge/push/etc? :)

Do note, this is not the same as `dune2themaker` (as that is in C++, etc). `dune2themaker4j` is a rewrite and a new engine written in `java`.

Awesome list you have there btw!

stefanhendriks:
oh I now checked and it looks like it does seem to be updated, however the status is still `halted`, could you plz change that into `Active` again? thx!

PR #362, opened 2016-10-09
Correct validation errors

wonea:


piranha:
There were two of those 'not playable', fixed in master.

PR #369, opened 2016-10-10
Client-side image validation

nikuda:
Use `OSGC.validate()` in the browser console to load and validate all image links from `games.yaml`

This is the result with the first 100 links, after the first 20-30 bad links were fixed:

screen shot 2016-10-10 at 7 12 03 pm

`OSGC.validate` takes 3 parameters:

* starting tick (where to start processing the array)
* how many ticks to do from the starting tick (default not set, does all)
* size of a tick (default 10, which is 10 image links in parallel every tick)

**Depends on** https://github.com/piranha/cyrax/pull/5

### DIFF
--- a/index.html
+++ b/index.html
@@ -136,7 +136,6 @@
       {{ game['name'] }}
     {%- if 'url' in game %}</a>{% endif %}
 
-
     {%- if 'repo' in game %} (<a href="{{ game['repo'] }}">repo</a>){%- endif %}
 
     {{ render_tag('status', game['status'] | title) }}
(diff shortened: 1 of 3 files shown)


nikuda:
Full run:

screen shot 2016-10-10 at 7 41 33 pm

index.html
```diff
@@ -146,8 +145,17 @@
...
+
+    {%- set images = [] %}
+    {% for item in game.get('media', []) %}
+      {% if 'image' in item -%}
+        {% do images.append(item['image']) %}
+      {%- endif %}
+    {% endfor -%}
     <div class="gallery" style="display: none"></div>
+    {% if images | length > 0 -%}
+      <script class="gallery-json" type="text/json">{{ images | join(',') }}</script>
```
nikuda: The images are not live DOM though, they're just plain text between `<script>` tags. They're made live when you click the image. So to collect them you'd have to put them in the DOM then read the `src` attribute. The plan is to remove all that and just use the new urls.

nikuda: Could revert back to that if you don't want the `do` extension in cyrax...

piranha: I was thinking about joining together `gallery-raw` and `gallery-json`, but I've looked into it and it's easier the way you did it.

PR #378, opened 2016-10-12
Missing repos added for a, b, w, x games

nikuda:


piranha:
It seems like a less gargantuan task right now, don't you think? :-)

PR #392, opened 2016-12-02
Add Stunt Rally as clone to TrackMania

Hiradur:


games/t.yaml
```diff
@@ -520,12 +520,32 @@
     - http://taisei-project.org/static/media/taisei_mansion.png
 
 - name: TrackMania
+  meta:
+    genre: [Driving/Racing]
+    subgenre: []
+    theme: []
```
nikuda: You can remove keys that don't hold useful information.

-subgenre, -theme

PR #394, opened 2016-12-03
Add Inexor as remake to Cube 2: Sauerbraten

Hiradur:


wonea:
Sorry not sure about this one, as Cube 2 was open source originally.

wonea:
Sorry, just realised Glest is added - so we'll have this as well.

PR #401, opened 2016-12-08
JA2 meta, C-Dogs YT, License formatting

nikuda:


games/c.yaml
```diff
@@ -385,7 +385,7 @@
     - http://raw.github.com/cxong/cdogs-sdl/master/wiki/images/thumbs/3.png
     - http://raw.github.com/cxong/cdogs-sdl/master/wiki/images/thumbs/4.png
     video:
-      youtube: fSIOdqZFM48
+      youtube: DysXfUyw58M
```
cxong: Why change the video back? That one is 3 years old

nikuda: The other one is in German and 2.5 years old..

Just noticed that you released a new trailer though, I'll put that one up.

PR #404, opened 2016-12-11
Add Harvest Cynthia as clone to Story of Seasons series (formerly Har…

Hiradur:
…vest Moon)

### DIFF
--- a/games/s.yaml
+++ b/games/s.yaml
@@ -760,6 +760,24 @@
     development: active
     lang: C#
     framework: Unity
+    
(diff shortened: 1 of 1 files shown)


Hiradur:
According to the project's readme very basic functionality already works but is buggy. Does this count as semi-playable already?

piranha:
Who knows, we don't have any strict rules about deciding on playability! Up to you to decide. :)

wonea:
Looks good to me

PR #408, opened 2016-12-29
The Battle for Wesnoth doesn't use Emacs Lisp

IBPX:
While it is listed on the GitHub repository, upon closer inspection, you can see the only files written in Lisp are Emacs modes intended for editing game data (or something along those lines). It isn't actually used in the game.


piranha:
:-))

PR #417, opened 2017-01-08
Genre improvements

nikuda:
It seems like RT and TB tactics games where merged into RTS/TBS. This reverts those back into use, along with genre updates.


wonea:
Thanks

PR #418, opened 2017-01-10
New fixed menu

nikuda:
Not having to scroll up to the tags clouds / filter helps

screen shot 2017-01-10 at 1 39 33 pm

Added filtering by "new", "added", or "updated".

screen shot 2017-01-10 at 1 39 09 pm

Cleaned up html, mobile view.

### DIFF
--- a/games/v.yaml
+++ b/games/v.yaml
@@ -32,6 +32,8 @@
       development: active
       status: playable
       updated: 2017-01-07
+      video:
+        youtube: IoF-kqXF2JQ
 
 - name: VVVVVV
   meta:
--- a/index.html
+++ b/index.html
@@ -11,6 +11,10 @@
   <title>Open Source Game Clones</title>
 </head>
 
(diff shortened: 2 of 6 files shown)


wonea:
I like it but want to try it out first. @piranha what do you think?

wonea:
Thanks! Is there any chance we could make the menu drop down sticky? Have a little pin icon on it?

nikuda:
Sounds good. I'm not 100% that this is the final solution but it's a step in the right direction. We could also merge it onto a new 'dev' branch and collectively improve it from there if it meets the bare minimum of improvement. 

I'll move 4b8ec92da3bd37823dfc6a3248347eb5348ef04e into a separate PR as that can be merged straight away, it cleans up a lot the HTML structure.

nikuda:
Ah never mind, should've refreshed this page before commenting. 

Think we can make the menu sticky with :focus. 🤔

Issue #416, opened 2017-01-07
SimCopter clone


wonea:
https://github.com/fr0zi/sccopter

Issue #415, opened 2017-01-07
Zork Clone


wonea:
https://github.com/vatbub/zorkClone

PR #421, opened 2017-01-10
New game additions

### LINKED ISSUE #414: Frontier First Encounters remake
https://en.wikipedia.org/wiki/Frontier:_First_Encounters

jffe
https://github.com/videogamepreservation/jjffe

GLFFE - Nic Mod
http://nic.dnsalias.com/FFE.html

wonea:
Fixes #414 #415 #416

### DIFF
--- a/games/e.yaml
+++ b/games/e.yaml
@@ -251,6 +251,28 @@
     images:
     - http://www.txori.com/data/images/articles/140205_frontier_ppsspp_android.png
 
+- name: [Elite III, 'Frontier: First Encounters']
+  names:
+  - "Frontier: First Encounters"
+  - "FFE"
+  meta:
+    genre: [Action, Simulation]
+    subgenre: [Flight Simulator]
+    theme: [Sci-Fi, Espionage]
+  remakes:
+  - name: jffe
+    updated: 2016-01-10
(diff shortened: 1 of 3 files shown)


OvermindDL1:
For note, sccopter is stated by the author to be a private project and not really for public use, being made for him to test things.  Also its state is minor reading of the sc2 file format and displaying of the tiles, it can do nothing else as of it, thus it should probably not be added.

cxong:
But it has an open source license; just need to make sure it doesn't have the playable tag.

OvermindDL1:
> But it has an open source license; just need to make sure it doesn't have the playable tag.

True, but he's implied it is mostly to learn the formats, not necessarily get something playable.  Although potentially someone could make something playable from it in the future once the formats are figured out.

Perhaps there needs to be a 'to-research' tag of some sort?

nikuda:
>Although potentially someone could make something playable from it in the future

FWIW, there are a quite a few of these early-stage projects in the data already.

wonea:
Thanks for the feedback, I've incorporated the changes.

Maybe status should have another entry for research.
status: # One of: research, unplayable, semi-playable, playable

PR #422, opened 2017-01-11
Sticky menu dropdowns

nikuda:
Per https://github.com/piranha/osgameclones/pull/418#issuecomment-271707909


wonea:
Great, this is now merged!

On a side note, do you think we could switch CSS theme around? I personally don't like black on white, and would rather it the other way round i.e. a dark theme. Maybe have this configurable? A little switch on the menu bar?

Issue #326, opened 2016-05-29
Grit Engine


Mailaender:
http://www.gritengine.com/ has asset extraction tools for Grand Theft Auto: San Andreas. Not sure if this counts as clone or reimplementation. https://github.com/sparkprime/grit-engine/tree/master/gtasa

wonea:
Fascinating, a bit like SCUMMVM but for open world games. Thanks for bringing this to our attention! I'm not entirely sure whether this go under GTA: San Andreas or all the other games listed on the Grit website; GTA series, Saints Row, Crysis, Fallout 3, Just Cause 2, Mafia 2, Burnout, etc.

Quite like this video:

https://www.youtube.com/watch?v=gARVzn3nBsQ

cxong:
This would be a tool; need to implement this first: #71

cxong:
This one is a bit tricky, because it also counts as a clone for the other games mentioned. For now I think we can just add it as an extraction tool for GTA:SA, since it doesn't seem like any of the other games are playable.

Issue #69, opened 2014-06-15
Settlers of Catan


neuromancer:
Despite it is a board game originally, we could include the free implementations of Settlers of Catan, I found 3 opensource implementations:
- Pionners
- JSettlers2
- Cities3d (code dumped in github)

PR #431, opened 2017-02-12
Added VASSAL engine, ordering change, utf-8 fixes

nikuda:


wonea:
Looks good, although Visual Pinball is actually a game. Shouldn't that remain so? I understand the logic behind VASSAL but Visual Pinball is different.

cxong:
no I don't think all board games / pinball games should be lumped together. Also I'm not so sure about including board games unless they are a clone of an existing video game. See #69

nikuda:
> Looks good, although Visual Pinball is actually a game. Shouldn't that remain so?

Was there a Visual Pinball game that Visual Pinball is an open source version of? I thought Visual Pinball was the original emulator of actual pinball machines and therefore doesn't have a parent game that it is a remake/clone of.

nikuda:
> no I don't think all board games / pinball games should be lumped together. 

Yea this will be a difficult one, it's a similar problem to ScummVM. Do we list all the games that ScummVM supports or do we just have it's entry as an exception to all other entries. I think ScummVM, VASSAL and Visual Pinball are the ones we should make an exception for until a better solution is found.

> Also I'm not so sure about including board games unless they are a clone of an existing video game.

Me neither, don't have a good answer to this but felt VASSAL might be ok since it's a not a new game but a digital version of existing games. Kind of like Visual Pinball.

cxong:
What about this: put these compilation games/engines into their own list, just as we currently have with remakes/clones?

On that note, we should have links at the top of the page to jump to these lists, maybe as a sidebar. The list is now so long I have trouble finding where the remakes list ends and the clones list begins.

wonea:
I agree with all these points;

- Create separate file compilation items, ScummVM and #326 has been waiting for a place.
- Clones and reimplementations should be better displayed together.

Also Visual Pinball was a game, and should remain listed under that.

https://en.wikipedia.org/wiki/Visual_Pinball

On a side note: wouldn't be cool if the first letter of a game and was listed down the side. Much like browsing through an address book?

[http://www.brianhadaway.com/wp-content/uploads/2011/05/ioslist2.png](url)

PR #451, opened 2017-04-20
some minor updates

Translator5:
fix offline pages
fix daggerfall name

### DIFF
--- a/games/d.yaml
+++ b/games/d.yaml
@@ -1,9 +1,9 @@
-- name: Daggerfall
+- name: Dark Forces
   names:
-  - Dark Forces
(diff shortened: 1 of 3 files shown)


cxong:
I just noticed the XL Engine entry. It's a general purpose engine that can run completely unrelated games. I don't think it makes sense to list them all together like we are doing right now, especially when there are other clones that only run Daggerfall. Each unrelated game should be listed separately, and XL Engine listed under that. The problem is that we'd have to duplicate the XL Engine entry, unless we update the data mapping between game and clone, and how the site is generated.

PR #447, opened 2017-04-19
Game corrections and additions

wonea:
Fixes #442 #445 #441 #438 #437 #448 #450

PR #452, opened 2017-04-21
Update GTA OpenRW repo URL

jaens:
Looks like the repo has moved? linking to the repo with the most activity


jaens:
Ah, apparently #447 fixes this as well.

PR #469, opened 2017-08-22
Removed Bejeweled clone and false alias of Exile

gakowalski:
- Repo for clone of Bejeweled is non-existent.
- Name "Blades of Exile" is not an alias of Exile (1988) or even game based on it but a different game from Exiles series by Spiderweb Software.


wonea:
Thanks

Issue #426, opened 2017-01-16
Dark Themed website


wonea:
It would be nice if osgameclones.com could optionally be presented with a dark theme. All my programming IDEs run dark themes; Visual Studio 2015 & Code, Notepad++, etc.

![2017-01-19 15_36_48-dark themed website issue 426 piranha_osgameclones](https://cloud.githubusercontent.com/assets/557192/22113108/2562eec4-de5d-11e6-8683-f68393bbbe0f.png)

JG916:
Just submitted a pull request for a dark theme. Let me know if you need any changes.

wonea:
@JG916 thanks for all the changes, is there any chance to change the CSS for tags to be something a little darker.

![image](https://user-images.githubusercontent.com/557192/31863331-3e0735fa-b744-11e7-8685-02c25cf9bc26.png)

@JG916 I've also suggested a change for CSS with the payable status #495

JG916:
Pull request submitted for the tag changes + question asked in the game status issue.

PR #473, opened 2017-10-01
Dark Theme #426

JG916:


wonea:
Thanks, its a good start

wonea:
Could we have the dark theme text indented hard right

image

The filters drop downs need the theme applied

image

Could we remember whether the theme has been applied, a cookie or session variable?

JG916:
Just made another pull request for the three fixes.

OvermindDL1:
> a cookie

:+1:  on this specifically

PR #499, opened 2017-10-23
change playable/semi-playable/unplayable tags background-color to gr…

JG916:
…een/orange/red and make text color darker in dark theme


JG916:
I had to change the html creation code a tiny bit in order to be able to use ".tag.playable" instead of ".tag-playable", but it seems to work without a problem now.

wonea:
Thanks for that!

Issue #501, opened 2017-10-24
Merge remake and clones


wonea:
Do you think we should merge remake and clones together? I would also like to list by date added or updated? @nikuda @cxong

cxong:
I think so. It will also help games which have multiple inspirations.
The parent game should act more like a label. You can click on it to quickly filter games that are clones or inspired by it.

wonea:
Okay, perhaps we should add a tag of 

type: [clone|remake] ?

cxong:
I've split the games YAML files to originals and clones. Now games can be clones/remakes of multiple games.
This is only a backend change though, in the front end the entries are repeated.

cxong:
Let's discuss how we'll present the list, now that the list is split and we can easily choose. We can show a list of originals plus their clones/remakes, or a list of all clones/remakes plus their original games. Or both?

For the combined originals list, if they have both clones and remakes, how should that be shown?

> Original game (RTS, Sci-Fi)
> - *CLONE* OpenGame (Playable, GPL, ...)
>    ...
>    ...
> - *REMAKE* SuperDuperGame (Not-Playable, Active, ...)
>    ...

?

wonea:
I think we should show both. Users struggle to know what the difference is, so I don't we should try to forcibly educate them.

cxong:
- [ ] Add verification for clones without remakes/clones
- [ ] Add verification for clones with non-existing original name

nikuda:
> type: [clone|remake] ?

I like this.

nikuda:
> I don't like the fact that we have two folders for clones and remakes now. Directory structure should go back to the simple listing of all, and the difference between remake/clones should be solved in the data structure/yaml.

To further the game itself is the top level, root node. By making the type the top node we introduce an odd way of thinking about it and there will be movements on the directory structure level. I.e. you change a type of game you move to it from one folder to the next and potentially delete/create new folders for new types. 

Everything except the game name is 'meta' information in terms of directory structure.

cxong:
We don't have two folders for remakes and clones; we have folders for originals and clones.

nikuda:
Same thing.

cxong:
I'm not sure what you mean. The current split is between originals and their clones/remakes. This is to make the schemas flexible enough to have games that clone or remake multiple originals, and is being used by one game already.

`type` doesn't make sense here, unless you mean to add `type: clone | remake` for the clones/remakes?

[...]

PR #510, opened 2017-10-29
Update OpenRCT2 Entry

Krutonium:
I updated the OpenRCT2 Entry to reflect the current status of the project. Is there a way, without copying and pasting of course, that I can indicate that the game is capable of replacing RollerCoaster Tycoon 1 as well? (Importing and running Scenarios/Saves/etc)

Additionally, we are technically a clone of "RollerCoaster Tycoon Classic" on iOS, Android, and Windows due to it being a re-release of the original game with a new UI on those platforms. We are also able to import this games assets as a base.

### DIFF
--- a/games/r.yaml
+++ b/games/r.yaml
@@ -219,12 +219,14 @@
   - name: OpenRCT2
     url: https://openrct2.com/
     repo: https://github.com/OpenRCT2/OpenRCT2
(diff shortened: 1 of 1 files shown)


cxong:
No; please see discussions in #501 for a possible solution to that

PR #534, opened 2017-12-12
Add platform key on originals, add games

### LINKED ISSUE #467: Supported Operating Systems
It would be useful if your YAML files could store information about supported OSes.
Not all game clones are available for all OSes.
Windows, OS X and Linux icons for each game clone would be useful.

nikuda:
Closes #467 and others. See commit messages.

### DIFF
--- a/schema/originals.yaml
+++ b/schema/originals.yaml
@@ -35,8 +35,10 @@ schema;platform:
   type: str
   enum: [
     'Amiga',
+    'Commodore 64',
     'Linux',
     'Mobile',
+    'MS-DOS',
     'Windows',
     'OSX',
   ]
--- a/schema/originals.yaml
+++ b/schema/originals.yaml
@@ -36,6 +36,7 @@ schema;platform:
   enum: [
     'Amiga',
     'Linux',
+    'Mobile',
     'Windows',
     'OSX',
   ]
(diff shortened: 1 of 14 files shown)


slx7R4GDZM:
https for Minetest links?

nikuda:
Sure, updated.

schema/originals.yaml
```diff
@@ -27,6 +31,21 @@ schema;meta:
...
+    'Apple II',
+    'Atari',
+    'Commodore 64',
+    'DOS',
+    'Linux',
+    'Mobile',
+    'MS-DOS',
+    'Windows',
+    'OSX',
+  ]
```
JayFoxRox: This platform scheme is slightly confusing.

Which version of Windows for example. There is a big difference between Windows 3.1 or Windows 10. Also OSX is not really precise either - I intend to soon add a game which works on the old Mac OS but not on newer macOS.

Also what's the difference between MS-DOS and just DOS? which version of a DOS?

Mobile is also kind of dangerous as Android also runs on TV sets etc. So putting Mobile can potentially break.

Either this should be more loose (e.g. device type this ran on to give people a rough idea) OR it should be more specific (so if people get rid of their old platform, they can look for alternative software).

It should also be considere to drop this field again  altogether (and some others too) and to replace it with some existing database / upstream (like information stored on wikipedia, mobygames or some other source).

cxong: Agree with some points, but this PR is already merged, so please create a new issue and continue this discussion there.

PR #536, opened 2017-12-29
Updated images of VCMI

sengaya:
Links taken from https://forum.vcmi.eu/t/some-screenshots-as-a-beginner-in-homm3-and-vcmi/1137


wonea:
Thanks

PR #546, opened 2018-02-22
Added OpenSAGE

Mailaender:
via https://forums.cncnz.com/topic/21241-first-version-of-opensage-published/

### DIFF
--- a/originals/c.yaml
+++ b/originals/c.yaml
@@ -85,6 +85,11 @@
     genre: [RTS]
     theme: [Sci-Fi, Modern Military]
 
+- name: Command & Conquer: Generals
+  meta:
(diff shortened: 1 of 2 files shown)


originals/c.yaml
```diff
@@ -85,6 +85,11 @@
     genre: [RTS]
     theme: [Sci-Fi, Modern Military]
 
+- name: Command & Conquer: Generals
```
cxong: I think yaml doesn't like the colon here, need to wrap in single quotes like here: https://github.com/Mailaender/osgameclones/blob/a3d6467770113543a8e2e6e8625d12d09f7612b5/originals/c.yaml#L129

PR #556, opened 2018-04-29
Update Hedgewars

nikuda:
Language corrections.

### DIFF
--- a/games/h.yaml
+++ b/games/h.yaml
@@ -73,7 +73,7 @@
   updated: 2017-10-23
   name: Hedgewars
   development: active
-  lang: [Lua, C, C++, Pascal]
+  lang: [Lua, C, C++, Pascal, Haskell]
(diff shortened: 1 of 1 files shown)


games/h.yaml
```diff
@@ -73,12 +73,14 @@
   updated: 2017-10-23
   name: Hedgewars
   development: active
-  lang: [Pascal, C++, Haskell]
+  lang: [Lua, C, C++, Pascal]
+  framework: [SDL, QT]
   repo: http://hg.hedgewars.org/hedgewars/
```
cxong: Why remove haskell? The gameserver uses it
http://hg.hedgewars.org/hedgewars/file/e461deddc942/gameServer

nikuda: My bad, it uses so many languages and I couldn't see Haskell code anywhere.

Issue #373, opened 2016-10-11
Content key for clones/remakes


nikuda:
New key in the clone/remake schema to indicate whether the game data is available

```yaml
name: freegish
repo: https://github.com/freegish/freegish
content: open
```

What could the acceptable values be? Maybe something like:

```yaml
content: commercial | free | open
```

piranha:
OpenTTD can work with original content and open source content, do we want to indicate that?

nikuda:
I added `swapable` to the enum list in #375. Can be changed later if it's not suitable..

Acceptable values:

```yaml
content: commercial | free | open | swapable
```

PR #572, opened 2018-06-16
add Shockolate

kparal:
This is a System Shock port based on source code for PowerPC
released by Night Dive Studios.

### DIFF
--- a/games/s.yaml
+++ b/games/s.yaml
@@ -162,6 +162,20 @@
   license: [As-is]
   remakes: [Shining Soul]
 
+- name: Shockolate
+  remakes: [System Shock]
+  repo: https://github.com/Interrupt/systemshock
+  development: active
+  status: playable
+  lang: [C]
+  framework: [SDL2]
+  license: GPL3
+  content: commercial
+  info: 'Needs game data from System Shock: Enhanced Edition'
+  updated: 2018-06-15
+  images:
+    - https://i.imgur.com/kbVWQj4.gif
+
 - name: Siege of Avalon Open Source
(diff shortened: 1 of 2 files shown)


games/s.yaml
```diff
@@ -162,6 +162,20 @@
   license: [As-is]
   remakes: [Shining Soul]
 
+- name: Shockolate
+  updated: 2018-06-15
+  repo: https://github.com/Interrupt/systemshock
+  development: active
+  status: playable
+  lang: [C]
+  framework: [SDL]
```
cxong: should be SDL2

kparal: I wasn't sure if frameworks are versioned and the readme didn't say. Fixed.

games/s.yaml
```diff
@@ -162,6 +162,20 @@
...
 
+- name: Shockolate
+  updated: 2018-06-15
+  repo: https://github.com/Interrupt/systemshock
+  development: active
+  status: playable
+  lang: [C]
+  framework: [SDL2]
+  license: GPL3
+  content: commercial
```
kparal: I am confused what `content` means. I assume it means game data as non-code files (graphics, music, etc). But does it describe what's **inside** the repo? Or what needs to be **supplied** (for remakes)? In Shockolate, there's no game content, because you need to supply original commercial files. Is that field value correct? (Also, could you explain this better in readme? Thanks).

cxong: See #373. It would be commercial

kparal: Ok. Anything else before this can be merged?

PR #578, opened 2018-06-20
add sfall (Fallout 2 remake)

kparal:


wonea:
Thanks!

cxong:
Not saying this shouldn't be added to osgameclones, but sfall is not like the other clones/remakes. It requires the original game software in addition to its data. Unlike a clone which requires original data, it will be much more difficult to produce replacement data for it. Therefore it is more like a patch or mod than a clone.

We should put more thought into how we can categorise such projects, including other open source patches or mods.

wonea:
Agreed, we've never added TDDPatch (which preceeded OpenTTD)

https://www.ttdpatch.net/
https://en.wikipedia.org/wiki/Transport_Tycoon#Third-party_creations

Issue #583, opened 2018-06-26
Combine remakes/clones lists, add more types


cxong:
We currently show separate lists for remakes and clones. For example this is what Spear of Destiny looks like:

![image](https://user-images.githubusercontent.com/1083215/41891907-99261262-7959-11e8-8f0a-82c761f0a91c.png)

Let's combine them like this:

![image](https://user-images.githubusercontent.com/1083215/41892001-f6f49dd2-7959-11e8-946f-b5fbcd860581.png)

Why? It looks neater, matches closer with how users might use the site (look for open source projects related to a game), and allows us to add more types of projects - tools #71, ports #533, patches (e.g. #580)

How to present these? I suggest turning it into a separate tag, like this:

![image](https://user-images.githubusercontent.com/1083215/41892194-92d2418c-795a-11e8-9410-a7de8538b85d.png)

tzizi:
To add support for more types of projects, I'm guessing a `type` attribute would be needed in the games folder.

Although it would require changing everything (possibly via a script?), this is the way I'd invision it:

```
    type: remake
    originals: [Bomberman]
```

Instead of the current:

```
    remakes: [Bomberman]
```

So then for tools or patches, it would simply be a matter of replacing the `type` attribute.

PR #587, opened 2018-07-24
Merge remakes and clones

tzizi:
Partially addresses #583. I also added a bit of wording in the index.html file to make the distinction between a remake and a clone clearer to the viewer.

### DIFF
--- a/_ext.py
+++ b/_ext.py
@@ -152,7 +152,8 @@ def parse_items(site, item, key):
             'framework',
             'content',
             'license',
-            'multiplayer'
+            'multiplayer',
+            'type'
         ]
 
         meta = item.get('meta', {})
@@ -280,17 +281,10 @@ def parse_data(site):
         combined = copy.deepcopy(item)
         name = game_name(combined)
 
-        combined_remakes = [
+        combined['games'] = [
             clone for clone in clones
-            if 'remakes' in clone and name in clone['remakes']
+            if name in clone['originals']
         ]
-
-        combined_clones = [
-            clone for clone in clones
-            if 'clones' in clone and name in clone['clones']
-        ]
-
-        combined['games'] = combined_remakes + combined_clones
         parse_items(site, combined, 'games')
 
 
--- a/_ext.py
+++ b/_ext.py
@@ -279,20 +279,19 @@ def parse_data(site):
         # Recombine originals and clones
         combined = copy.deepcopy(item)
         name = game_name(combined)
+
         combined_remakes = [
             clone for clone in clones
             if 'remakes' in clone and name in clone['remakes']
         ]
-        if len(combined_remakes) > 0:
-            combined['remakes'] = combined_remakes
+
         combined_clones = [
             clone for clone in clones
             if 'clones' in clone and name in clone['clones']
         ]
-        if len(combined_clones) > 0:
-            combined['clones'] = combined_clones
-        combined['both'] = combined_remakes + combined_clones
-        parse_items(site, combined, 'both')
+
+        combined['games'] = combined_remakes + combined_clones
+        parse_items(site, combined, 'games')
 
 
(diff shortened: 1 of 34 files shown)


nikuda:
I'll have to check this one out from home. Not sure if the python code is the clearest way of doing this.

nikuda:
Do you mind rebasing from master? It seems like a lot of the changes here have already made it into master from the other PRs you had. Thanks for the clean up by the way 👍

tzizi:
Sorry, I'm a bit new to this. I tried rebasing, but github's diff doesn't seem to work?

nikuda:
You need to be able to pull in changes from `opengaming:master` to your local repo. 

```
git remote add upstream git@github.com:opengaming/osgameclones.git
```
then this should do it

```
git fetch upstream
git rebase upstream/master combine-remakes-clones
git push
```

tzizi:
Thanks! Fixed :)

_ext.py
```diff
@@ -182,8 +189,8 @@ def sort_key(game):
         ]
         if len(combined_clones) > 0:
             combined['clones'] = combined_clones
-        parse_items(site, combined, 'remakes')
-        parse_items(site, combined, 'clones')
+        combined['both'] = combined_remakes + combined_clones
```
cxong: use something like 'games' instead of 'both' - after all, one day this would include more things? So 'both' would be inaccurate 😉

_ext.py
```diff
@@ -182,8 +189,8 @@ def sort_key(game):
         ]
         if len(combined_clones) > 0:
             combined['clones'] = combined_clones
```
cxong: 'clones' and 'remakes' aren't needed, since only 'both' is used

index.html
```diff
@@ -72,19 +72,15 @@ <h1>Open Source Game Clones</h1>
...
 
   <div id="content">
-    <h2>List of remakes</h2>
-    <article>Some of these games aren&#39;t exact remakes but evolution of original ones, which were eventually open sourced.</article>
+    <h2>List of games</h2>
+    <article>
+      A <span class="tag" data-name="remake">Remake</span> is a game where the executable and sometimes the assets as well are remade open source. Some of these games aren&#39;t exact remakes but evolution of original ones, which were eventually open sourced.
+      <br />
+      A <span class="tag" data-name="clone">Clone</span> is a game which is very similar to or heavily inspired by a game or series.
+    </article>
```
cxong: 👍

tzizi:
Thanks, fixed :)

What is your point of view on using the `type` attribute I suggested in #583 ? It would make the code much cleaner I believe, something like this:

```
combined['games'] = [
    clone for clone in clones
    if name in clone['originals']
]
```

And the tags hack could be eliminated:

```
game_tags = [
    ...
    'type'
]
```

With, of course, games.html being updated as well.

nikuda:
There have been discussions about having the `type: clone | remake` attribute for a while now, and I'd say everyone's in agreement that's the way it should be done - but no-one has gone and done it yet. PR for it would be welcome 👍

tzizi:
Gone and done it :)

Any ideas for the hover background for the Remake/Clone tags?

tzizi:
Rebased again, the commits look fine.

However, I found a bug with the tags. When clicking on the "Clone", if you scroll down to "Microsoft Train Simulator" (just an example, there are many others), you'll find OpenRails as expected, but also ICBM3D, which is a clone of Missile Command, repeated twice (for some reason, it appears to be duplicated, also seen on the website before this PR).

I'm not sure if this is a bug intrinsic in the tag system, or if I introduced it with this PR. I'll have to look into it.

tzizi:
Interesting, rebased again, however now it appears the Missile Command issue still exists (although the Super Methane Brothers one is now fixed)

Issue #589, opened 2018-07-24
Question about asdf


tzizi:
I'm currently working on an implementation for #520, but I found asdf (https://github.com/TimPietrusky/asdf) in the list of remakes that don't have an original game associated.

Does this count as an open-source game clone? It seems more like an electronic toy than a computer game to me. If it does, what should it be classified under, Arcade? Also, wouldn't this be considered as a clone, rather than a remake?

nikuda:
Yes it's a clone by our categorisation. Arcade seems like the most appropriate genre.

I'd add Simon https://en.wikipedia.org/wiki/Simon_(game) to the originals list and then link it to asdf.

PR #590, opened 2018-07-24
Validate original games

### LINKED ISSUE #520: Validate `remakes` and `clones`
For clones, `remakes` and `clones` refer to the original game. This needs to be validated, that the original exists, otherwise the clone won't show up in the output.

This can't be done using pykwalify; will need to check in python code.

tzizi:
Fixes #520, however, 2 games remain broken:

asdf, as mentioned in #589, and Scramble, since I have no idea what the game is, and the sources are down.

### DIFF
--- a/_ext.py
+++ b/_ext.py
@@ -110,17 +110,29 @@ def parse_items(site, item, key):
         getattr(site, key).append((names(item), meta, map(parse_fn, item[key])))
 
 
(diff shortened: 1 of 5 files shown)


tzizi:
Fixed both, I'm guessing it'll be a little complicated to merge with #587 though, maybe this one should be merged first, and #587 can be rebased over?

PR #591, opened 2018-07-25
Add repository icons

### LINKED ISSUE #514: Repo type icons
I would like some indicator for the type of repo, like an icon for example:

screen shot 2017-11-02 at 11 19 57 pm

I prefer github because it's easy to use for me, but others might prefer providers like bitbucket or gitlab for various reasons. Also I think it'll be great to see at a glance whether the repo link is actually pointing to a .zip or .tar.gz archive.

So far most repos are github, but we also have sourceforge, bitbucket, launchpad, google code, savannah, and zip/tgz blobs.

The icons may need to have dark versions.

tzizi:
Fixes #514.

The sources for the icons are as follows:

 * Github - from their press kit
 * Launchpad, Bitbucket, Sourceforge - website's favicons
 * Archive - Adwaita icon theme
 * Google - Wikipedia


_ext.py
```diff
@@ -81,9 +82,38 @@ def parse_global_tags(site, item, tag):
...
-    return dict(entry,
-                new=(date.today() - updated) < timedelta(days=30),
-                tags=parse_tags(entry, entry_tags) + parse_tags(meta, meta_tags))
+    result = dict(entry,
+                  new=(date.today() - updated) < timedelta(days=30),
+                  tags=parse_tags(entry, entry_tags) + parse_tags(meta, meta_tags))
+
+    if "repo" in result:
+        domain = re.sub(r".*://([^/]*)/.*", "\\1", result["repo"])
+        ext = re.sub(r".*?(\.[^/]*)$", "\\1", result["repo"])
```
cxong: use os.path.splitext

_ext.py
```diff
@@ -81,9 +82,38 @@ def parse_global_tags(site, item, tag):
...
     updated = entry.get('updated') or date(1970, 1, 1)
-    return dict(entry,
-                new=(date.today() - updated) < timedelta(days=30),
-                tags=parse_tags(entry, entry_tags) + parse_tags(meta, meta_tags))
+    result = dict(entry,
+                  new=(date.today() - updated) < timedelta(days=30),
+                  tags=parse_tags(entry, entry_tags) + parse_tags(meta, meta_tags))
+
+    if "repo" in result:
+        domain = re.sub(r".*://([^/]*)/.*", "\\1", result["repo"])
```
cxong: use urllib.parse

_ext.py
```diff
@@ -81,9 +82,38 @@ def parse_global_tags(site, item, tag):
...
+            result["repotitle"] = "Google Code"
+        elif "bitbucket.org" in domain:
+            result["repoicon"] = "bitbucket-32x32.png"
+            result["repotitle"] = "Bitbucket"
+        elif "launchpad.net" in domain:
+            result["repoicon"] = "launchpad.png"
+            result["repotitle"] = "Launchpad"
+        elif (".zip" in ext or
+              ".tar." in ext or
+              ".tgz" in ext):
```
cxong: ext in ('.gz', '.tgz', '.zip')

static/google.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 48 48"><defs><path id="a" d="M44.5 20H24v8.5h11.8C34.7 33.9 30.1 37 24 37c-7.2 0-13-5.8-13-13s5.8-13 13-13c3.1 0 5.9 1.1 8.1 2.9l6.4-6.4C34.6 4.1 29.6 2 24 2 11.8 2 2 11.8 2 24s9.8 22 22 22c11 0 21-8 21-22 0-1.3-.2-2.7-.5-4z"/></defs><clipPath id="b"><use xlink:href="#a" overflow="visible"/></clipPath><path clip-path="url(#b)" fill="#FBBC05" d="M0 37V11l17 13z"/><path clip-path="url(#b)" fill="#EA4335" d="M0 11l17 13 7-6.1L48 14V0H0z"/><path clip-path="url(#b)" fill="#34A853" d="M0 37l30-23 7.9 1L48 0v48H0z"/><path clip-path="url(#b)" fill="#4285F4" d="M48 48L17 24l-4-3 35-10z"/></svg>
```
cxong: Please provide licensing information for all these icons. We need to make sure that we have permission to use these logos in this way. The github logo is probably fine, with the condition that we can't modify the logos. What about the other logos? Are there licensing or usage terms?

How do these icons look in dark mode?

cxong:
👍  but as noted, it's important to make sure we have permission to use these icons.

static/google.svg
tzizi: Good point, this is why I noted the sources above, as I'm not sure myself.

The Github logo and the archive icon should be fine ([license](https://github.com/GNOME/adwaita-icon-theme/blob/master/COPYING)). The Google logo is from Wikipedia, [said to be in public domain, but with possible restrictions](https://commons.wikimedia.org/wiki/File:Google_%22G%22_Logo.svg).

The rest, I wasn't unable to find any official (usable) icons, so I used their favicons as a proof-of-concept. Should I just remove them for now?

slx7R4GDZM:
What about GitLab?

tzizi:
Added, icon is from [their press kit](https://about.gitlab.com/press/).

static/google.svg
cxong: Thanks for providing the archive icon and google logo icons. Please include this information in the code, maybe in the readme or `LICENSE` file, so that we don't lose track.

Note that the archive icon is CC-BY-SA, which requires attribution. We will need to do something like include a footer in the web page mentioning that this icon is CC-BY-SA licensed, and the author/source.

I'm not sure if it's ok to take favicons in general. It would be best to find ones that are clearly licensed.

Have you tried using a web font instead, like FontAwesome? They have icons for many popular web services - GitHub, Google, Bitbucket - and they can support light/dark versions too. This would make it much simpler.

nikuda: > Have you tried using a web font instead, like FontAwesome? They have icons for many popular web services - GitHub, Google, Bitbucket - and they can support light/dark versions too. This would make it much simpler.

This would be much preferred.

tzizi:
Replaced with FontAwesome, dark theme works as well. Still need to credit, but I'm not quite sure how it should be done.

cxong:
According to https://fontawesome.com/v4.7.0/license/ attribution is not required, so we don't have to do anything.

cxong:
According to https://fontawesome.com/license/free we don't need to do anything additional.

static/bitbucket-brands.svg
```diff
@@ -0,0 +1 @@
+<svg aria-hidden="true" data-prefix="fab" data-icon="bitbucket" class="svg-inline--fa fa-bitbucket fa-w-16" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"><symbol id="bitbucket" viewBox="0 0 512 512"><path fill="currentColor" d="M23.1 32C14.2 31.9 7 38.9 6.9 47.8c0 .9.1 1.8.2 2.8L74.9 462c1.7 10.4 10.7 18 21.2 18.1h325.1c7.9.1 14.7-5.6 16-13.4l67.8-416c1.4-8.7-4.5-16.9-13.2-18.3-.9-.1-1.8-.2-2.8-.2L23.1 32zm285.3 297.3H204.6l-28.1-146.8h157l-25.1 146.8z"></path></symbol></svg>
```
cxong: I think it's better to use CSS as suggested in https://fontawesome.com/how-to-use/on-the-web/setup/getting-started?using=web-fonts-with-css
Include their CSS via their CDN, and show the icons using `<i>` elements.

tzizi:
Good point, less code and renders better as well.

tzizi:
Probably best to leave this to another PR, but I'm guessing we should replace the RSS icon as well

cxong:
Sure; consistent icon styles is better

_ext.py
```diff
@@ -81,9 +82,37 @@ def parse_global_tags(site, item, tag):
...
+            result["repoiconstyle"] = "fab"
+            result["repotitle"] = "Google Code"
+        elif "bitbucket.org" in domain:
+            result["repoiconname"] = "bitbucket"
+            result["repoiconstyle"] = "fab"
+            result["repotitle"] = "Bitbucket"
+        elif "gitlab.com" in domain:
+            result["repoiconname"] = "gitlab"
+            result["repoiconstyle"] = "fab"
+            result["repotitle"] = "Gitlab"
```
slx7R4GDZM: Not that it matters since it's already been merged but the Lab in GitLab is capitalized.

tzizi: Thanks! #595

PR #594, opened 2018-07-26
Replace RSS icon with FontAwesome

tzizi:
Depends on #591 of course


static/main.css
```diff
@@ -190,6 +190,23 @@ dd + dt { margin-top: 1em; }
...
+  width: 100%;
+  display: inline !important;
+  padding: .2em;
+}
+
+.feed-icon:hover {
+  border-bottom: 0;
+  background-color: #f26522;
+  color: white;
+}
```
cxong: I'm not sure if custom styling is needed. We don't need to keep the original 14x14 size, ideally it would be inline with the text. Not sure if those colours will work in both light and dark mode.

PR #603, opened 2018-08-01
Update The Settlers I, II & III details

hellcp:
Some missing details, original names, American release names, fixed images, added links.

### DIFF
--- a/originals/t.yaml
+++ b/originals/t.yaml
@@ -194,7 +194,7 @@
     genre: [Platform]
 
 - name: [The Settlers, The Settlers (video game)]
-  names: [The Settlers II]
+  names: [Serf City]
   meta:
     genre: [RTS]
 
--- a/originals/t.yaml
+++ b/originals/t.yaml
@@ -76,11 +76,22 @@
   meta:
     genre: [Action, Adventure]
 
-- name: The Settlers II
+- name: [The Settlers, The Settlers (video game)]
+  names:
+  - [Die Siedler II]
+  - [Serf City]
   meta:
     genre: [RTS]
 
-- name: The Settlers III
+- name: [The Settlers II, The Settlers II]
+  names:
+  - [Die Siedler II]
+  meta:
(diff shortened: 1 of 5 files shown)


originals/t.yaml
```diff
@@ -76,13 +76,26 @@
   meta:
     genre: [Action, Adventure]
 
-- name: The Settlers II
+- name: [The Settlers, The Settlers (video game)]
+  names: [Die Siedler II, Serf City]
```
cxong: is this right? The Settlers is also called "Die Siedler II"?

originals/t.yaml
```diff
@@ -76,13 +76,26 @@
...
+    subgenre: [City Building]
+    theme: [Medieval, Alternate Historical, Fantasy]
 
-- name: The Settlers III
+- name: [The Settlers III, The Settlers III]
+  names: [Die Siedler III]
   meta:
     genre: [RTS]
+    subgenre: [City Building]
+    theme: [Medieval, Egyptian, Alternate Historical, Fantasy]
```
cxong: I don't think `Medieval` is right here, from the description it seems more like mythological, and for time period antiquity or ancient history. Unfortunately the Giant Bomb themes aren't the best. It would be good to come up with a better set of themes later.

originals/t.yaml
wonea: It's right
https://www.mobygames.com/game/die-siedler-ii-die-nchste-generation-wikinger

originals/t.yaml
wonea: Shall we create Medieval? For now?

originals/t.yaml
hellcp: That's a good catch, it's Die Siedler and not Die Siedler II

originals/t.yaml
hellcp: I would have used Ancient if that was a theme, unfortunately it isn't :/

Issue #406, opened 2016-12-11
Not fair to call some games clones? Needs discussion.


Hiradur:
I realized I may have been a little bit to quick with adding clones. All I wanted is to spread awareness for some FOSS games. However, Duion, the Uebergame developer told me on IRC he sees Uebergame as an original game, not a clone of ARMA or anything else. He is fine with keeping it in the list for promotion for now though.
However, this made me think. Some of the games I added look pretty similar to the originals (to me) but since I didn't play most of them or didn't play them long enough they may not actually be clones or don't want to be considered clones by their developers/community.

Games in question are:
- Annex: Conquer The World: Does not list CnC as inspiration or anything but the game looks very similar, one community member said that listing it as a clone is probably fine and could make Annex more popular
- Bridge Command (seems to be a very complex simulator, not sure how it compares to the Ship Simulator series)
- Stunt Rally (doesn't look exactly like TrackMania but IMO it definitely has some elements from it)
- Rigs of Rods (BeamNG.drive actually is a commercial project started by the original Rigs of Rods devs after RoR. The focus of the games seems to be slightly different. Rigs of Rods strives to be a hardcore simulator while BNG wants to be a bit more accessible I think (don't hold me on that, I haven't played it and don't want to start a flamewar ;) ). Being a former Rigs of Rods developer I can say that there have always been tensions and flamewars between the two communities but the games look similar in some aspects)
- Spring 1944: Lists both Company of Heroes and Supreme Commander as an inspiration so calling it a clone of CoH may not be right (although I wrote about CoH and SC being inspirations in the info section)
- The Butterfly Effect (this one does look _very_ similar to Crazy Machines and The Incredible Machine so listing it as a clone may be fine although these games are nowhere listed as an inspiration)

What do you think should be done about this? Should I ask the developers of these games if they approve their games being listed as clones? I don't want to do any damage, e.g. that people join the project with wrong expectations. Or should these games rather be moved to the loosely related games category which I proposed in #398? IMO this would be the best option.

cxong:
For obvious clones there's no problem. When there's doubt, I would get a second opinion, from the developers or from players very familiar with the games. Then when there's still doubt, we can look at them on a case-by-case basis.

For uebergame, I bet the *real* objection is the negative connotation of the word "clone". So I think we should avoid using that word altogether, and renaming the section "List of games inspired by" or something like that. "Clones" may be factually accurate but the word has baggage.

foobar13373:
[...]

PR #607, opened 2018-08-04
Remove Rigs of Rods

ZentroPlays:
Rigs of Rods is a not a clone of BeamNG. Did the person who add this bother to lookup the history between these two games? Rigs of Rods was released in 2005. BeamNG was officially launched in 2015 by former developers of Rigs of Rods.

### DIFF
--- a/games/r.yaml
+++ b/games/r.yaml
@@ -240,31 +240,6 @@
   updated: 2015-04-03
   url: http://sourceforge.net/projects/rickyd/
 
-- name: Rigs of Rods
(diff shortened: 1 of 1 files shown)


cxong:
Please join the earlier discussion at #406

wonea:
Agreed, I would rather it marked as something than removing it entirely.

cxong:
I've updated the ror entry here: https://osgameclones.com/#beamng.drive-games
Please close if this is acceptable

PR #608, opened 2018-08-06
Add link to space nerds in space webpage

smcameron:
Signed-off-by: Stephen M. Cameron <stephenmcameron@gmail.com>

### DIFF
--- a/games/s.yaml
+++ b/games/s.yaml
@@ -461,8 +461,9 @@
   originals:
   - 'Artemis: Spaceship Bridge Simulator'
   repo: https://github.com/smcameron/space-nerds-in-space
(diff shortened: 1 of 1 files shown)


games/s.yaml
```diff
@@ -461,8 +461,9 @@
   originals:
   - 'Artemis: Spaceship Bridge Simulator'
   repo: https://github.com/smcameron/space-nerds-in-space
-  type: clone
-  updated: 2015-05-02
+  type: remake
```
cxong: the type should be something else; remakes are for things like source ports, or engine reimplementations where the original assets can be used.

PR #614, opened 2018-08-06
Change type of "Space Nerds In Space" from "clone" to "inspired"

smcameron:
Clone means: genetically identical
http://lmgtfy.com/?q=define%3Aclone

SNIS shares zero DNA with Artemis, and is merely the same genre,
so it is cannot be a clone. Calling it a clone would be like calling
Star Wars a clone of Star Trek merely because they both have spaceships.

Signed-off-by: Stephen M. Cameron <stephenmcameron@gmail.com>

### DIFF
--- a/games/s.yaml
+++ b/games/s.yaml
@@ -461,7 +461,7 @@
   originals:
   - 'Artemis: Spaceship Bridge Simulator'
   repo: https://github.com/smcameron/space-nerds-in-space
-  type: clone
+  type: inspired
   updated: 2018-08-05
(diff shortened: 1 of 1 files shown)


cxong:
Inspired is a good fit; we need to add it to the schema too though. Let's see if other members agree with the new tag in #406

wonea:
It's got my vote.

cxong:
I've changed it from "inspired" to "similar". Based on the discussion here #406, having too many word choices is redundant. If necessary we can use the `info` field to provide more context.

PR #625, opened 2018-10-05
Updated Commander Keen series

pehsa:
The Keen Dreams (2015) releases has been removed from Steam, Itch.io and IndieGameStand. The repository is for the open sourced original game.
Added the Reflection Keen Dreams, which is based on the original source, but the development active (last update 5 months old)


cxong:
5 months is a long time, not sure I agree with active status. Maybe sporadic?

pehsa:
I will change the status at night.

PR #627, opened 2018-10-11
Fix repo link of zaz

basxto:


games/z.yaml
```diff
@@ -42,7 +42,7 @@
   development: halted
   originals:
   - Zuma
-  repo: https://zaz.svn.sourceforge.net/svnroot/zaz
+  repo: https://sourceforge.net/p/zaz/code/HEAD/tree/
```
cxong: I think just having the repo as `https://sourceforge.net/projects/zaz/` is fine.
I've gone over lots of other entries, seems like many sourceforge repo links are broken because they've changed their routes or the project has changed VCSs which also changes the route. We can avoid these by only linking to sourceforge.

PR #629, opened 2018-10-23
Add Mininim

wviana:
Is it all it needs?

### DIFF
--- a/games/m.yaml
+++ b/games/m.yaml
@@ -396,7 +396,6 @@
     - 91QCqPMssj4
     - HI5dl5sENsw
 
-
 - name: Mirror Magic
   lang: C
   license:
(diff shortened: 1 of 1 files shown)


games/m.yaml
```diff
@@ -377,6 +377,25 @@
...
+  status: playable
+  content: open
+  originals:
+  - Prince of Persia
+  repo: https://github.com/oitofelix/mininim
+  type: remake
+  updated: 2017-01-12
+  url: http://oitofelix.github.io/mininim/
+  video:
+    youtube:
```
cxong: only one video supported at the moment

wviana: Updated it as `youtube: 91QCqPMssj4` is that it?

PR #632, opened 2018-11-02
Chronoshift

OmniBlade:
Adds the Chronoshift project which recently went public with its source.

Also Updates Thyme entry to indicate it is playable.

### DIFF
--- a/games/c.yaml
+++ b/games/c.yaml
@@ -272,6 +272,18 @@
   type: remake
   url: https://www.chocolate-doom.org/
 
+- name: Chronoshift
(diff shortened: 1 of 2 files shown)


stavrossk:
How do you guys know that Chronoshift is playable if it hasn't any binary release?

OmniBlade:
There are CI builds available, depends what is needed to count as playable, the project piggybacks on the original binary so even though it doesn't implement enough code by itself to be playable, it injects in the original and replaces its code to form a playable release.

PR #642, opened 2018-12-14
Add Dust Racing 2D

cxong:


wonea:
I like the mud effects!

wonea:
Okay to merge in.

PR #650, opened 2019-01-18
Racer halted (fixes #649)

### LINKED ISSUE #649: Racer no longer active?
Racer is marked as active, however the last commit on SF was 4 years ago, so I think this may be incorrect.

cxong:


nikuda:
Can we hide the DangerJS bot when it has nothing to report? 'Files in PR' is already under GitHub's 'Files changed' tab.

cxong:
It's a work in progress; eventually it will show more useful information, like games added/changed. Will need to add YAML parsing first.

nikuda:
Oh I see. I thought it coming through only on errors, but I can see other bits of information being useful.

PR #679, opened 2019-03-21
Add SuperTuxParty

pehsa:


originals/m.yaml
```diff
@@ -110,6 +110,15 @@
   meta:
     genre:
     - Racing
+    
+- name: Mario Party
+  external:
+    wikipedia: Mario Party
+  meta:
+    genre: 
+    - Sports
```
cxong: hmm I'm not completely sold on our current genres; it says it's based on giant bomb but there they don't have the genre/subgenre split, and Mario Party is listed as "Trivia/Board Game". I feel like the genre should be "Party", and we should go with wikipedia's classification system: https://en.wikipedia.org/wiki/Template:Video_game_genre

Going to discuss this in a separate issue.

cxong:
```suggestion
    - Party
```

PR #686, opened 2019-03-24
Add Riiablo

pehsa:


pehsa:
I asked the developer which license the project use, becasue no license information on the project page.

cxong:
You can use `As-is` in the meantime

PR #687, opened 2019-03-25
Add Future Blocks

pehsa:


pehsa:
I deleted the duplicated repo tag for the FreeDroid. I don't know why the failed check

pehsa:
The problem is FreeDroidRPG and FreeDroidClassic not the same game. FreeDroidClassic is the original remake of the ParaDroid. Top-Down Shoot'em Up game.

FreeDroidRPG, we could say, is a full rewrite and now it is a Paradroid themed, isometric game with many RPG elements.

pehsa:
I don't know whats the real problem with these. Earlier I found this error, but closed the Pull.

In the original file, FreeDroidRPG had 2 repo tag. I removed one, but the Travis build log says when checking the original f.yaml theres a problem. Not able to check the modified file, because the original is bad. I don't know how to resolve this.

PR #688, opened 2019-03-25
Add SwtPacMan

pehsa:


games/s.yaml
```diff
@@ -1045,6 +1045,23 @@
...
+  originals:
+  - Pac-Man
+  repo: https://github.com/fogleman/SwtPacMan
+  url: https://www.michaelfogleman.com/projects/pac-man/
+  development: halted
+  status: playable
+  lang: Java
+  framework: SWT
+  license:
+  - MIT
```
cxong: I can't find this license in the project pages; where is it?

pehsa: My bad! Michael Fogleman's other projects (Craft, Future Blocks, etc.) released with MIT license. I don't checked this project. I change it to As-is

PR #714, opened 2019-04-18
Add Achtung, die Kurve! (kurve.se)

SimonAlling:
This is, I believe, the most faithful existing clone of the original game from 1995. Thanks to @jancborchardt for suggesting that it be added to OSGC.

### DIFF
--- a/games/a.yaml
+++ b/games/a.yaml
@@ -51,6 +51,31 @@
   updated: 2016-12-11
   url: https://achtungkurve.com/
 
+- name: Achtung, die Kurve!
+  images:
+  - https://kurve.se/screenshots/kurve-round.png
+  - https://kurve.se/screenshots/kurve-splash.png
+  - https://kurve.se/screenshots/kurve-lobby.png
+  info: The original <kbd>ZATACKA.EXE</kbd> in the browser
+  framework:
+  - HTML
+  - CSS
+  lang:
+  - JavaScript
+  license:
+  - AGPL3
+  development: complete
+  originals:
+  - Achtung, die Kurve!
+  repo: https://github.com/SimonAlling/kurve
+  content: open
+  status: playable
+  type: clone
+  multiplayer:
+  - Split-screen
+  updated: 2019-04-18
+  url: https://kurve.se/
+
 - name: AdaDoom3
   lang: Ada
   license:
--- a/games/a.yaml
+++ b/games/a.yaml
@@ -35,20 +35,20 @@
   - https://achtungkurve.com/dist/images/screenshot_1.png
   - https://achtungkurve.com/dist/images/screenshot_2.png
   - https://achtungkurve.com/dist/images/screenshot_3.png
-  framework:
-  - HTML
-  - CSS
   lang:
   - JavaScript
   license:
   - GPL3
-  development: sporadic
+  development: active
   originals:
   - Achtung, die Kurve!
   repo: https://github.com/maechler/kurve
   status: playable
-  type: remake
-  updated: 2016-12-11
+  type: clone
+  multiplayer:
+  - Competitive
+  - Local
+  updated: 2019-04-29
   url: https://achtungkurve.com/
 
 - name: Achtung, die Kurve!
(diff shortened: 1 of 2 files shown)


jancborchardt:
Can confirm it’s the most faithful to the original. No weird items, no line smoothing, KONEC HRY – I love it. :+1: :smile:

games/a.yaml
```diff
@@ -51,6 +51,31 @@
...
 
+- name: Achtung, die Kurve!
+  images:
+  - https://kurve.se/screenshots/kurve-round.png
+  - https://kurve.se/screenshots/kurve-splash.png
+  - https://kurve.se/screenshots/kurve-lobby.png
+  info: The original <kbd>ZATACKA.EXE</kbd> in the browser
+  framework:
+  - HTML
+  - CSS
```
cxong: these aren't really frameworks; this should be removed

games/a.yaml
```diff
@@ -51,6 +51,31 @@
...
+  - AGPL3
+  development: complete
+  originals:
+  - Achtung, die Kurve!
+  repo: https://github.com/SimonAlling/kurve
+  content: open
+  status: playable
+  type: clone
+  multiplayer:
+  - Split-screen
```
cxong: hmm our schema breaks down here; it's shared screen competitive gameplay. Not sure what we should do here

SimonAlling: Me neither. I tried to pick the best one out of the available options. I don’t know exactly what to do, but I know one thing for sure: It must say multiplayer one way or another.

games/a.yaml
SimonAlling: Agree. Only reason I included those was because I copied the existing _Achtung, die Kurve!_ entry (by maechler) and modified it.

games/a.yaml
cxong: Well since we're using giantbomb for genres, I checked and here's their multiplayer features list:

<select id="multiplayer" name="multiplayer[]" multiple="multiple"><option value="25">Asynchronous multiplayer</option><option value="16">LAN co-op</option><option value="19">LAN competitive</option><option value="15">Local co-op</option><option value="18">Local competitive</option><option value="21">Local splitscreen</option><option value="17">Online co-op</option><option value="20">Online competitive</option><option value="22">Online splitscreen</option><option value="23">Pass and play</option><option value="24">Voice chat</option></select>

So I guess we can add "Competitive" to our schema

games/a.yaml
SimonAlling: Should I make this change, `git commit --amend` and then `git push --force` (to the `kurve` branch of my fork)?

maechler: @SimonAlling Actually I did not add the game myself initially, but I just checked the initial commit when my implementation was added, there it said: `lang: [Javascript, CSS, HTML]`
I do not why this was moved to `framework`.

maechler:
@SimonAlling I do not about any definitions for those words, but I would probably also go for `remake` for `achtungkurve.com`. 

Also I thought that probably the two games should not both be called `Achtung, die Kurve!`? 
If there is an alternative name for `kurve.se` that would be fine for me, else we could rename mine to `achtungkurve.com` or probably better `Kurve`.

SimonAlling:
Yeah, @maechler, the name problem has been bogging me for quite a while, even before I even knew about OSGC. On the one hand, I want the name of my clone to indicate my intention that it be as close as possible to the original, and I didn't have any better idea than simply _Achtung, die Kurve!_. On the other hand, a unique name would be beneficial in many situations.

maechler:
@SimonAlling I would not mind renaming my game to `achtungkurve.com` or `Kurve`, especially since yours is really the one closest to the original.

SimonAlling:
@maechler That sounds great to me if it's OK with you. 😃

cxong:
@SimonAlling I didn't think too hard about the remake/clone definition, but I think it's worth revisiting. Usually we've used remake for games that either use the original source, or can load the original assets - like an engine reimplementation. But that's not really a definition useful to players. The key difference is what kind of experience the player should expect. For remakes they should expect more or less the same game, whereas for clones the link is much looser.

For most other games it was easy as you just needed to look at the source and whether it can use original assets. For Achtung, die Kurve! it's a bit trickier, as it has no assets. So it'll take people familiar with the gameplay to be able to judge fairly. Perhaps both can be remakes in this case.

maechler:
@cxong Thanks for your explanation! After reading this I would probably go for `clone` for `achtungkurve.com` and `remake` for `kurve.se`. But then `Zatacka X` should probably also be considered `clone`, not sure about  `Netacka`.

PR #715, opened 2019-04-18
Add shell.nix

SimonAlling:
It's convenient for Nix(OS) users, who only have to run

    nix-shell

to get a sandbox with all dependencies available.

PR #716, opened 2019-04-18
Add .venv/ to .gitignore

SimonAlling:
It showed up under "Untracked files", and it seems to me that it shouldn't be committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _build/
 .env/
+.venv/
 node_modules/
 *.pyc


cxong:
By the way this is only the case when you use pipenv with `PIPENV_VENV_IN_PROJECT=1`

SimonAlling:
> By the way this is only the case when you use pipenv with `PIPENV_VENV_IN_PROJECT=1`

I see. As can be seen in #715, I use a `shell.nix` file which sets that environment variable. I got it from a guy over at the `#nixos` channel at Freenode; I couldn't build OSGC until I added it.

PR #721, opened 2019-05-01
add freeaoe

sandsmark:


games/f.yaml
```diff
@@ -261,6 +261,25 @@
   updated: 2014-04-22
   url: https://freeablo.org/
 
+- name: freeaoe
+  development: active
+  lang:
+  - C++
```
```suggestion
  - C++
  framework:
  - SFML
```

sandsmark: github doesn't seem to understand that I included your fix already

PR #723, opened 2019-05-04
Add OutpostHD

pehsa:


games/o.yaml
```diff
@@ -1422,6 +1422,28 @@
   updated: 2015-04-15
   url: http://identicalsoftware.com/ostrichriders/
 
+- name: Outpost HD
+  type: remake
```
cxong:
```suggestion
  type: inspired
```
Looks like it's aiming for a different experience than the original, so this is more appropriate.
From the forum post:
> OutpostHD is a remake of OUTPOST by Sierra On-line from 1994. It's not a clone of OUTPOST -- it's an entirely new game inspired by OUTPOST's core mechanics while providing a more complete and satisfying experience than the original ever could.

games/o.yaml
```diff
@@ -1422,6 +1422,28 @@
...
+  originals:
+  - Outpost
+  repo: https://github.com/ldicker83/OPHD
+  url: https://forum.outpost2.net/index.php/topic,5718.0.html
+  development: sporadic
+  status: playable
+  license:
+  - BSD
+  content: free
```
cxong:
```suggestion
  content: commercial
```
Looks like they haven't replaced the original assets yet. I'm not sure if the game is freeware yet, so this should be commercial

games/o.yaml
```diff
@@ -1422,6 +1422,28 @@
   updated: 2015-04-15
   url: http://identicalsoftware.com/ostrichriders/
 
+- name: Outpost HD
+  type: remake
+  originals:
+  - Outpost
+  repo: https://github.com/ldicker83/OPHD
+  url: https://forum.outpost2.net/index.php/topic,5718.0.html
+  development: sporadic
```
cxong:
```suggestion
  development: active
```

PR #732, opened 2019-05-08
Dungeon craft: Updated repository

GaidamakUA:
Replaced old Sourceforge repository with new GitHub one.

### DIFF
--- a/games/d.yaml
+++ b/games/d.yaml
@@ -681,10 +681,10 @@
   development: active
   originals:
   - 'Forgotten Realms: Unlimited Adventures'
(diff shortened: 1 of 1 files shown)


GaidamakUA:
Thank you, that's reasonable.

wonea:
@cxong @GaidamakUA Should we reference the original and new version being worked on? Seems the new fork is worth of addition to the site.

GaidamakUA:
@wonea unfortunately, I won't be able to work on the project.
@cxong I think, it's reasonable to close pull request.

PR #740, opened 2019-05-14
Add NBlood

pehsa:


games/n.yaml
```diff
@@ -13,6 +13,22 @@
...
+  originals:
+  - Blood
+  repo: https://github.com/nukeykt/NBlood
+  url: https://nukeykt.retrohost.net/
+  development: active
+  status: playable
+  lang:
+  - C++
+  license:
+  - GPL
```
cxong:
```suggestion
  - GPL2
```

pehsa: My bad! I checked the EDuke32 license on the frontpage. They say the project using GPL license, but the link is for the GPL2 license.

Issue #742, opened 2019-05-14
Fix GPL games


cxong:
Usually games use GPL2 or GPL3. There is a GPL version 1 but it is very rare; most games currently listed as GPL are most likely misclassified.

- Check all the games currently listed as GPL, change as necessary: https://github.com/opengaming/osgameclones/search?q=GPL&unscoped_q=GPL
- Add a Danger check if the license is GPL

tzizi:
Maybe `GPL` could also be renamed to `GPL1` to make it clearer to both users of the site, and people creating pull requests?

cxong:
Similar thing with LGPL actually - there is no version 1, so it should be removed altogether

cxong:
Hopefully having it called `GPL1` will prevent people from using it by mistake

PR #744, opened 2019-05-15
Fixing GPL license

pehsa:
I do not find GPL version specification for 3 games:
- Interstate Outlaws
- Duck Rehunt
- rawgl

### DIFF
--- a/games/g.yaml
+++ b/games/g.yaml
@@ -62,7 +62,7 @@
 - name: Geometry Wars Remake
   lang: Java
   license:
-  - Artistic
+  - As-is
   development: halted
   originals:
   - Geometry Wars
--- a/games/g.yaml
+++ b/games/g.yaml
@@ -62,7 +62,7 @@
 - name: Geometry Wars Remake
   lang: Java
   license:
-  - GPL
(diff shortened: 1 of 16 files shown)


pehsa:
#742

games/g.yaml
```diff
@@ -62,7 +62,7 @@
 - name: Geometry Wars Remake
   lang: Java
   license:
-  - GPL
+  - Artistic
```
cxong: I'm not sure about this license; this repo is in google code so it's not easy to check, but is there an actual license file? I wonder why it would be listed as "Artistic License/GPL" on the project page. It could be some default value, like the first option in a dropdown.

pehsa: Yeah, the Google Code default is Artistic/GPL, but no GPL version specification on the site. I think the Artistic license for the assetts and GPL for the code...

cxong: If the license files cannot be found, this should be As-is

pehsa: Changed the Artistic licenses to As-is and modified the 3 games too, which was unable to determine the version.

PR #746, opened 2019-05-17
Add Cave Story Engine 2

pehsa:


games/c.yaml
```diff
@@ -173,6 +173,26 @@
...
+  
+- name: Cave Story Engine 2
+  type: remake
+  originals:
+  - Cave Story
+  repo: https://github.com/cuckydev/Cave-Story-Engine-2
+  development: active
+  status: playable
+  lang:
+  - C
```
```suggestion
```
github detects header files as C but I think the project is pure C++, as shown in the makefile

PR #750, opened 2019-05-21
Add Grimsonland

pehsa:


games/g.yaml
```diff
@@ -278,6 +278,23 @@
...
+  repo: https://github.com/BrooklynKing/Grimsonland
+  url: http://brooklynking.github.io/Grimsonland/
+  development: halted
+  status: playable
+  lang:
+  - JavaScript
+  framework:
+  - Phaser
+  license:
+  - As-is
```
cxong:
```suggestion
  - MIT
```
See https://github.com/BrooklynKing/Grimsonland/blob/master/package.json#L10

PR #769, opened 2019-06-06
Mark "Hedgewars" as playable

utdemir:
It is playable even more so than the original, with decent amount of active games on any time of the day.

### DIFF
--- a/games/h.yaml
+++ b/games/h.yaml
@@ -130,7 +130,7 @@
   status: playable
   repo: http://hg.hedgewars.org/hedgewars/
   type: clone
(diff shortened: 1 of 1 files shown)


games/h.yaml
```diff
@@ -127,6 +127,7 @@
   development: active
   originals:
   - Worms Series
+  status: playable
   repo: http://hg.hedgewars.org/hedgewars/
   type: clone
   updated: 2017-10-23
```
cxong:
```suggestion
  updated: 2019-06-07
```

PR #770, opened 2019-06-06
Add proxx

augbog:
proxx is a game unveiled at Google I/O which is a Minesweeper clone. It's available for all platforms including desktop and mobile.

A few things I wasn't entirely sure was for "framework", proxx actually allows you to specify different rendering whether it be with WebGL or canvas (see README in repo). I'm not sure if it makes sense to mention both but willing to fix anything if necessary.

### DIFF
--- a/games/p.yaml
+++ b/games/p.yaml
@@ -438,6 +438,22 @@
   updated: 2015-04-03
   url: http://quakeone.com/proquake/
 
+- name: proxx
+  development: complete
+  images:
+  - https://user-images.githubusercontent.com/234957/57080494-8caff680-6cea-11e9-8477-54ea0e1670e2.png
+  repo: https://github.com/GoogleChromeLabs/proxx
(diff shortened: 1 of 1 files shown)


games/p.yaml
```diff
@@ -438,6 +438,24 @@
...
+  - https://user-images.githubusercontent.com/234957/57080494-8caff680-6cea-11e9-8477-54ea0e1670e2.png
+  repo: https://github.com/GoogleChromeLabs/proxx
+  url: https://proxx.app/
+  license:
+  - Apache
+  status: playable
+  framework: WebGL
+  lang:
+  - TypeScript
+  - JavaScript
```
cxong:
```suggestion
```
looks like the only js files in the project are config/packaging, so it's not as important to mention them

augbog: Ah fair point thanks will fix that :)

PR #781, opened 2019-06-09
Add Q-Gears (Final Fantasy VII)

pcwalton:


games/q.yaml
```diff
@@ -1,3 +1,17 @@
+- name: Q-Gears
+  images:
+  - http://q-gears.sourceforge.net/ss.jpg
+  lang: C++
+  license:
+  - GPL2
+  development: halted
+  originals:
+  - Final Fantasy VII
```
cxong: This game will need to be added in the originals

games/q.yaml
```diff
@@ -1,3 +1,17 @@
+- name: Q-Gears
+  images:
+  - http://q-gears.sourceforge.net/ss.jpg
+  lang: C++
```
cxong:
```suggestion
  lang: C++
  framework: OGRE
```

games/q.yaml
```diff
@@ -1,3 +1,17 @@
...
+  license:
+  - GPL2
```
cxong:
```suggestion
  - GPL2
content: commercial
```

games/q.yaml
wonea: Sorted

Issue #147, opened 2015-04-08
Enemy Engaged repository?


wonea:
Enemy Engaged was released as open source but have to join mailing list to get access to their GIT server. Unsure whether they want their GIT address public. Any comment?

http://eechcentral.simhq.com/index.php?title=Development

https://en.wikipedia.org/wiki/Enemy_Engaged:_RAH-66_Comanche_vs._KA-52_Hokum

Mantis - Enemy Engaged
http://bugs.enemyengaged.net/my_view_page.php

nikuda:
This original source release. We don't have a way of marking it as such. I'd keep holding the addition until we have a way of designating original source releases from remakes/clones.

nikuda:
Blocked by #533

enemy-engaged-developers:
This repo isn't open source, please don't mark it as such.

PR #798, opened 2019-06-20
Add Enemy Engaged (#147)

cxong:


wonea:
Noticed some C++ code in there too.

cxong:
The C++ code might be part of 3rd party libraries only

Issue #774, opened 2019-06-07
Enduro


rafaelcastrocouto:
New game
https://racascou.itch.io/enduro

```yaml
- name: Enduro Tribute
  type: remake
  originals: [Enduro]
  repo: https://github.com/rafaelcastrocouto/enduro
  url: https://racascou.itch.io/enduro
  development: complete
  status: playable
  lang: [HTML5, Javascript, CSS]
  license: MIT
  content: open
  info:
  updated: 5/19
  images:
    - https://img.itch.zone/aW1nLzIxMzUyMzgucG5n/original/DZBGLA.png
```

cxong:
https://en.wikipedia.org/wiki/Enduro_(video_game)

PR #820, opened 2019-07-11
Add Enduro Tribute (#774)

cxong:


wonea:
@cxong Just spotted the licence is wrong should be 'As is'

Issue #833, opened 2019-07-24
Add gorillas-rs


cxong:
https://github.com/nodef0/gorillas-rs

PR #840, opened 2019-08-02
Add Gorillas-rs

pehsa:
#833

### DIFF
--- a/games/g.yaml
+++ b/games/g.yaml
@@ -249,6 +249,23 @@
   updated: 2015-04-06
   url: https://github.com/Lyndir/Gorillas
 
+- name: Gorillas-rs
+  type: clone
+  originals:
+    - Scorched Earth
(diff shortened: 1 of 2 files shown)


games/g.yaml
```diff
@@ -249,6 +249,23 @@
   updated: 2015-04-06
   url: https://github.com/Lyndir/Gorillas
 
+- name: Gorillas-rs
+  type: clone
+  originals:
+    - Scorched Earth
```
cxong: the original should be gorilla.bas
https://en.wikipedia.org/wiki/Gorillas_(video_game)

pehsa: Changed! Sorry, I don't know the original game. The gameplay was similar to Scorched Earth and that was my first thought.

cxong:
@nodef0 FYI

PR #854, opened 2019-08-17
Corrected spelling of some game names

Trilarion:
I corrected the spelling of some game names. Check the URLs of the game entries to see that the spelling is now the correct one for the entries.

Is the update field the update of the game entry in the database or the update (last release date) of the game? If the former, I would still need to update it.

### DIFF
--- a/games/c.yaml
+++ b/games/c.yaml
@@ -36,7 +36,7 @@
   video:
     youtube: hgX3jz_HajE
 
-- name: C-Evo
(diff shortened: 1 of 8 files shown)


Trilarion:
I see from the bot output that the update field should be updated. Will do it.

Trilarion:
Updated fields are updated now.

wonea:
Thanks!

Issue #824, opened 2019-07-17
Add Tetris (in C and ncurses)


cxong:
https://github.com/brenns10/tetris
https://brennan.io/2015/06/12/tetris-reimplementation/

polyzen:
http://victornils.net/tetris/
https://github.com/vicgeralds/vitetris

PR #864, opened 2019-08-27
Add tetris (#824)

cxong:


wonea:
Unusual find! I wonder if they're any PowerShell games out there!

Issue #671, opened 2019-03-15
Optimise site for mobile


cxong:
Site starts zoomed-in and has a huge right margin in mobile

pehsa:
Closable #885

PR #885, opened 2019-09-04
Modify min and max width

pehsa:
I changed the body max-width value and added min-width. This will make the site more comfortable with almost every screen size:

1920x1080:
kép

1280x960:
kép

iPad (768x1024) (from Firefox Dev Tools):
kép

iPhone 6/7/8 (375x667) (from Firefox Dev Tools):
kép

iPhone X/XS and Galaxy S9/S9+ similar to iPhone 6/7/8.

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -16,7 +16,7 @@ header, section, footer, aside, nav, article, figure { display: block; }
 
 /* Markup */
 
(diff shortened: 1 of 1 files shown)


cxong:
#671

PR #887, opened 2019-09-05
applied automatic alphabetic sorting of the yaml entries

Trilarion:
(see script at https://gist.github.com/Trilarion/34ebcdf9d50323ab47f75382318bf793)


Trilarion:
Any idea why dangerJS crashed? Maybe because the sorting moved entries from one yaml file to another?

pehsa:
yaml.reader.ReaderError: 'utf8' codec can't decode byte #xd6: invalid continuation byte

  in "/home/travis/build/opengaming/osgameclones/games/d.yaml", position 10342

The problematic game entry is DonkeyInTrouble. The LÖVE's "Ö" character. I think this will crash with the other 2 LÖVE games (mari0 and EXILE)

Trilarion:
I can confirm that the encoding is not UTF-8 as I wanted. May be a problem with ruamel.yaml. Need to check.

games/l.yaml
```diff
@@ -79,78 +79,6 @@
   type: remake
   updated: 2017-10-15
 
-- name: The Legend Of Zelda - Navi's Quest
```
cxong: we want to ignore common words like "The" when ordering
https://en.wikipedia.org/wiki/Alphabetical_order#The_and_other_common_words
See https://github.com/opengaming/osgameclones/blob/master/_ext.py#L216-L223 for an example of how to handle this
This is so we don't end up with a ton of entries in `t.yaml` for example

Trilarion: Good idea. I will implement it in a further version. This will move entries currently in t.yaml like The Battle for Wesnoth somewhere else.

cxong:
Unfortunately we're still on python 2.7; will need to update cyrax to python 3 first.

pehsa:
When this will be happen? :D

I saw you tried to make a Python3 version in last year, but nothing happened yet.

There will be good if we don't rely on Python2 because the Linux distros now shipping default with Python3, and often cause problems when Python2 and 3 installed on the same system. I have one separate VM just for the OSGC Python2 dep.

cxong:
it's so long I don't remember. There's a PR here https://github.com/piranha/cyrax/pull/9 but I guess extra testing is required, and for @piranha to take a look

Trilarion:
I got ruamel.yaml to load and dump yaml as UTF-8 (see for example DonkeyInTrouble, which leaves "LÖVE" as "L\xD6VE". I used the RoundTripLoader and Dumper which leaves the structure as is mostly, which is well suited for a mix of manual programmatic editing of the files. I ignored "the" and "a" in front of titles, resulting in "The Battle for Wesnoth" now in b.yaml and so on. I tested the script on Python 2 and 3 and it works with both and gives the same results. The updated script is in the gist, use it if you want.

However, dangerJS in the build still crashes with the "Cannot read property '1' of null" error.

Trilarion:
It may be that dangerJS doesn't support moving a game from one yaml file to another.

pehsa:
Not just that. I checked the DangerJS and I think every moving type unsupported. It's check the games before modify and after modify and compare the metadata for changes. But you just moved the entries and not modifying the data. The games not really changed and tried to check those modified entries. 
For example the checkRepoAdded got NULL value for the game and tried to check game.repo.

One solution if you want to pass the checks: Not just sort the games, but if moved than change the updated field.

Trilarion:
There is a logic error here. There is no game added, just moved but DangerJS thinks games have been added.

Second thing is that this is not really a game update, just a sorting. Why should the update fields change if nothing really changed? What we need is a check that all the content has indeed remained the same, that no game has been added or removed and no content has been changed. Then we could let the script run occasionally. Like a maintenance task.

pehsa:
The DangerJS chack the git diffs. The diffs says the games changed. The JS load those games the before and after states and check the changed values, and store the really changed data. But here no really changed data.

Trilarion:
I wonder if it is rather a latent error that got only discovered now. It could be because only the modified entries are checked.

Error:  TypeError: Cannot read property '1' of null
    at checkRepoAdded (dangerfile.js:43:58)

looks like 

  if (game.repo && game.repo.indexOf('github.com') >= 0) {
    const author = game.repo.match(/github.com\/(\w+)\//)[1]

The check that the repo contained 'github.com' was successful, but the regex match was unsuccessful.

Trilarion:
I thought that the repo for "Yamagi Quake II", which is https://github.com/yquake2 might have caused the crash in dangerJS, but maybe it wasn't the only one. Might need to run dangerJS myself.

games/y.yaml
```diff
@@ -5,9 +5,9 @@
   development: active
   originals:
   - Quake 2
-  repo: https://github.com/yquake2
+  repo: https://github.com/yquake2/yquake2
```
cxong:
```suggestion
  repo: https://github.com/yquake2
```
let's leave this unchanged; if dangerJS is broken it should be fixed instead

games/y.yaml
```diff
@@ -5,9 +5,9 @@
...
   type: remake
-  updated: 2015-04-05
+  updated: 2019-09-07
```
cxong:
```suggestion
  updated: 2015-04-05
```
Since nothing really changed, leave the updated date as-is
Don't worry about dangerJS warnings, they are warnings only and meant for game additions/updates

Trilarion:
Thanks for updating the dangerJS file. Requested changes are done. From my side it's ready to be merged.

PR #894, opened 2019-09-18
Cleanup

Zehra:
Repo is gone because of DMCA takedown request.


wonea:
Always sad to see things go...

PR #905, opened 2019-09-30
corrected writings of frameworks

Trilarion:
HTML Canvas, Gradle, ZLIB, boost are not real frameworks
unified way of writing pygame, melonJS, LibGDX, ..


wonea:
Thanks

Issue #903, opened 2019-09-27
Warn on other unusual field content


Trilarion:
Analogously to #786 (which can be closed now) we could also check and warn for other known field contents. For the various proposals please see the posts below. If there is a consensus, I could make a PR for it.

Trilarion:
The development field has the following content (number of occurrences in parentheses)

halted(326), active(325), sporadic(124), complete(32), very active(6)

Therefore we could use these 5 words as valid development attributes and check for them.

Trilarion:
The type field has the following content:

remake(670), clone(165), similar(10), tool(4)

We could use these 4 words as valid type attributes and check for them.

Trilarion:
The multiplayer field has the following content:

Online(28), LAN(16), Competitive(9), Split-screen(7), Local(3), Hotseat(3), Co-op(2)

We could use these 7 words as valid multiplayer field attributes.

Trilarion:
The content field has the following content:

commercial(95), open(59), free(27), swappable(5)

We could use them as valid content attributes.

Trilarion:
The status field has the following content:

playable(263), semi-playable(31), unplayable(30)

We could use them as valid status field attributes.

Trilarion:
Regarding the multiplayer field, should it maybe also be converted to start with lowercase (except for LAN) to be more in line with the other field contents?

Trilarion:
Additionally, fields license and framework could be checked.

cxong:
these are enforced by the schema check already and will fail the build
danger is only used for warnings, where there is uncertainty whether a strict schema is required
except for framework, we could do that.

Trilarion:
Frameworks are quite a lot:

[...]

PR #910, opened 2019-10-02
known frameworks check in danger

Trilarion:
see discussion in #903


Trilarion:
demo worked (misspelled a framework), will remove it

wonea:
@cxong @Trilarion Any chance we could move the frameworks and languages into the schema folder?

cxong:
I don't think we should; the schema is very strict and will prevent the site from building. We should let people add new languages and frameworks freely.

PR #937, opened 2019-10-17
Add github stars badge

cxong:


cxong:
image

Added lazy loading to avoid hitting API limits (http code 429)

cxong:
Looks like the meta badges should be moved up a bit, so that the bottoms align with the bottom of text

PR #946, opened 2019-10-20
add Zero-K

mzemb:


games/z.yaml
```diff
@@ -103,6 +103,24 @@
   video:
     youtube: RKmKpTe-gVo
 
+- name: Zero-K
+  type: remake
```
cxong:
```suggestion
  type: similar
```

games/z.yaml
```diff
@@ -103,6 +103,24 @@
...
+  lang:
+  - C#
+  license:
+  - GPL
```
cxong:
```suggestion
  - GPL2
```

games/z.yaml
```diff
@@ -103,6 +103,24 @@
   video:
     youtube: RKmKpTe-gVo
 
+- name: Zero-K
+  type: remake
+  lang:
+  - C#
```
cxong:
```suggestion
  - C#
  - Lua
```

games/z.yaml
```diff
@@ -103,6 +103,24 @@
...
+  license:
+  - GPL
+  development: active
+  originals:
+  - Total Annihilation
+  - Supreme Commander
+  status: playable
+  repo: https://github.com/ZeroK-RTS
+  url:
```
cxong: url only supports one entry

games/z.yaml
```diff
@@ -103,6 +103,24 @@
...
+  originals:
+  - Total Annihilation
+  - Supreme Commander
+  status: playable
+  repo: https://github.com/ZeroK-RTS
+  url: 
+   - https://zero-k.info/
+   - https://store.steampowered.com/app/334920/ZeroK/
+  video:
+   youtube: dyhzqcf7NJM
```
cxong:
```suggestion
    youtube: dyhzqcf7NJM
  updated: 2019-10-21
```

games/z.yaml
mzemb: I think both entries are equally important here. :/

PR #971, opened 2019-11-08
Update h.yaml

VictorJL:
Harvest Moon project remake made for UWP and Xbox One

### DIFF
--- a/games/h.yaml
+++ b/games/h.yaml
@@ -70,7 +70,7 @@
   originals:
   - Story of Seasons series
   repo: https://github.com/greentwip/HarvestMoon
-  status: playable
+  status: semi-playable
   type: remake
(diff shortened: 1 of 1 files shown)


games/h.yaml
```diff
@@ -59,6 +59,22 @@
...
+  info: Harvest Moon (SNES) remake for UWP and Xbox One.
+  lang:
+  - C#
+  license:
+  - GPL3
+  development: active
+  originals:
+  - Story of Seasons series
+  repo: https://github.com/greentwip/HarvestMoon
+  status: playable
```
cxong:
```suggestion
  status: semi-playable
```
Looks like gameplay features are missing compared to the original, so this is more appropriate

PR #982, opened 2019-11-14
Add ReTux

cxong:


wonea:
Why so many licenses?

cxong:
The CC licenses and GPL2 are for the content; see http://git.savannah.nongnu.org/cgit/retux.git/tree/data/LICENSES

Issue #949, opened 2019-10-21
Update stargus repo


cxong:
https://github.com/Wargus/stargus

PR #1014, opened 2019-12-15
Update stargus #949

cxong:


wonea:
The YouTube video shows it's very buggy.

PR #1043, opened 2020-01-05
add TekwarGDX

CommonLoon102:
Related issue: https://github.com/opengaming/osgameclones/issues/1038

### DIFF
--- a/games/t.yaml
+++ b/games/t.yaml
@@ -113,7 +113,7 @@
   framework:
   - LibGDX
   license:
-  - GPL3
+  - Custom
   updated: '2020-01-05'
   video:
     youtube: 0U7EpJlf6-o
--- a/games/t.yaml
+++ b/games/t.yaml
@@ -82,6 +82,7 @@
   type: remake
   updated: 2019-08-22
   url: https://qnighy.github.io/tapir/
+
 - name: Terasology
   lang:
   - Java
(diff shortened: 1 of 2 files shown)


games/t.yaml
```diff
@@ -98,6 +99,26 @@
   video:
     youtube: cU6LL5BXi5E
 
+- name: TekwarGDX
+  originals:
+  - TekWar
+  type: remake
+  repo: 'https://gitlab.com/m210/TekwarGDX'
+  url: 'http://m210.duke4.net/'
+  development: very active
```
cxong:
```suggestion
  development: active
```

games/t.yaml
```diff
@@ -98,6 +99,26 @@
...
+  status: playable
+  content: commercial
+  lang:
+  - Java
+  framework:
+  - LibGDX
+  license:
+  - GPL3
```
cxong: This doesn't look GPL3 compatible
https://gitlab.com/m210/TekwarGDX/blob/master/buildlic.txt
```suggestion
  - Custom
```

games/t.yaml
```diff
@@ -98,6 +99,26 @@
...
+  info: 'Works on Linux too, using OpenJDK 8'
```
cxong:
```suggestion
```
don't think this is needed

games/t.yaml
CommonLoon102: In that case, Chocolate Duke3D, Duke3D, Duke3d_win32, EDuke32, JFDuke3D, xDuke, NBlood, PCExumed and Rednukem are not GPL2 compatible either, because they have this file too.

cxong: https://en.wikipedia.org/wiki/Build_(game_engine)#Other_game_ports
yeah seems that's true, good to know

CommonLoon102: OK, so what to do? Have here `Custom ` and then change it to `Custom ` for every other Build game in a separate PR?

cxong: yep that sounds good

PR #1066, opened 2020-02-07
add SR tool

kparal:
That includes creating 'Septerra Core: Legacy of the Creator' original
entry and adding 'UFO: Enemy Unknown' alternative name (not directly
related to SR, I just noticed the world-wide name was missing).

### DIFF
--- a/games/s.yaml
+++ b/games/s.yaml
@@ -394,7 +394,7 @@
   updated: 2019-11-15
   video:
     youtube: 6v1USVhtfjQ
-  
+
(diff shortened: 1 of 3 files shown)


wonea:
@cxong I'm happy to add this, would do you think?

kparal:
For example for Albion, there are some enhancements (in addition to a native binary) mentioned here:
https://github.com/M-HT/SR/blob/master/games/Albion/release/linux/readme-Linux.txt

PR #1076, opened 2020-03-22
Added my project Wolf3dX for Wolfenstein 3D

francot514:
Added the project template


francot514:
Ok my bad it will requiere game files, but can work with shareware version

PR #1085, opened 2020-05-01
Add dRally

mnhauke:


cxong:
I'm not sure we should add this; the source code is mostly disassembled so cannot be open source.

neuromancer:
I think it should be accepted, since we already included Devilution or Alive Reversing.

PR #1090, opened 2020-05-10
Another clone of minecraft

Holmes4657:


games/m.yaml
```diff
@@ -386,6 +386,26 @@
...
+  status: playable
+  lang:
+  - C
+  framework:
+  - OpenGL
+  license:
+  - MIT
+  updated: '2020-05-10'
+  images:
+  - 'https://github.com/jdah/minecraft-weekend/blob/master/screenshots/1.png'
```
cxong, line 405: this link doesn't lead to an image file; normally you need to click the Raw button to get to the image link but I see that this file is too big so github serves it as a download instead.

wonea:
Sorry the images are still broken

'https://raw.githubusercontent.com/SourceCodeDeleted/world-of-might-and-magic/master/install_images/gameplay.PNGS

Holmes4657:
This is not my part of the added code. This is added by itself when I edit the code. Weird

PR #1097, opened 2020-05-19
Updated Ultimas, Empire, and Xconq.

Feneric:
Expanded out the names of the Ultima games that weren't already expanded and fixed all the references to them.
Added two Ultima games not already listed along with the open-source games they've influenced.
Added Empire.
Added Xconq.


games/x.yaml
```diff
@@ -1,3 +1,20 @@
+- name: Xconq
+  images:
+  - https://en.wikipedia.org/wiki/Xconq#/media/File:Xconq.jpg
```
cxong, line 3: this should link to an image

Feneric: Oops, sorry about that. Fixed.

PR #1101, opened 2020-05-20
Another clone of minecraft

Holmes4657:


Holmes4657:
I deleted the previous one as some mistakes were made

PR #1106, opened 2020-06-07
Add original Driver 2 Playstation and game project REDriver2

francot514:
Add the original game to list and the open source remake project.

### DIFF
--- a/games/r.yaml
+++ b/games/r.yaml
@@ -164,6 +164,24 @@
   video:
     youtube: KSoVdF9SvNY
 
+- name: REDriver2
+  type: remake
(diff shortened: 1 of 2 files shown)


games/r.yaml
```diff
@@ -164,6 +164,24 @@
...
+  development: active
+  status: playable
+  lang:
+  - C++
+  framework:
+  - SDL
+  - SDL2
+  - OpenGL
+  license:
+  - Custom
```
cxong, line 181: I can't find a license in their repo; if it's not specified then the license should be `As-is`

PR #1122, opened 2020-08-23
Update OpenSupaplex

vrza:
OpenSupaplex is a 1:1 reimplementation of the original Supaplex game in C and SDL, with just a few changes to make it easy to use with game controllers, and also easy to port to any platform.

### DIFF
--- a/games/o.yaml
+++ b/games/o.yaml
@@ -1612,8 +1612,8 @@
   - Supaplex
   repo: https://github.com/sergiou87/open-supaplex
   status: playable
-  type: remake
-  updated: 2020-05-06
+  type: clone
+  updated: 2020-08-24
   images:
   - https://i.imgur.com/6FUB9tO.png
   - https://i.imgur.com/5OK0Yld.png


cxong (changes requested):
opensupaplex already exists; please update the existing copy, keeping things alphabetically sorted

vrza:
Updated.

Seeing that OpenBOR immediately follows Open BattleChess makes it unclear how strings with space characters are sorted. A linting tool that checks the sort order in the CI workflow seems like the way to go.

cxong (changes requested):
I think remake is more appropriate for open supaplex

vrza:
Do we have these categories (remake, clone...) defined somewhere?

cxong:
It's pretty much the blurb you see at the top of the webpage
In some cases the definitions are a bit fuzzy, but I think for opensupaplex, since the goal is a 1:1 reimplementation, a remake is more appropriate than clone.

vrza:
The OpenSupaplex README says:

the main goal of OpenSupaplex is having a 1:1 _clone_ of the original Supaplex game, in every possible aspect, and only making exceptions wherever is reasonable to do so

Technically, OpenSupaplex was made by rewriting assembly code obtained by disassembling the DOS executable into C and SDL, subtroutine for subroutine. The label that comes to my mind first is "source port" -- it preserves the original program structure, uses the same format for loading assets etc.

Looking at that blurb, "remake" is defined as "a game where the executable and sometimes the assets as well are remade open source".

This is a bit different than the definition of "Video game remake" on Wikipedia, which reads:

"A remake offers a newer interpretation of an older work, characterized by updated or changed assets."

( https://en.wikipedia.org/wiki/Video_game_remake )

But here we are using the label "remake" also for source ports that use the original assets verbatim.

Anyway, by the current definition on this website "remake" is the best-fitting label for OpenSupaplex, so we can close this PR.

Issue #93, opened 2014-11-03
Sort by added/updated


SamHasler:
Could it be possible to sort the list by added/updated date so that when someone returns to the page they can see what was added since they last visited.

cxong:
I've been thinking about this issue, and I think it might be best to start looking into a web framework like React or Vue. We currently do our tags/filters using basic javascript; it works but is not the most maintainable. With a framework this kind of stuff can be added easily. Thoughts?

piranha:
With framework like that you'll either have only client-side rendering or a need for a much more elaborate setup with server-side rendering, instead of simply serving an HTML file like it works right now.

Maybe it's better to look instead at some frameworks which on one hand make code a little bit more structured, but on the other do not require client side rendering?

Maybe (stimulus)[https://github.com/stimulusjs/stimulus]? It's what RoR uses IIRC.

cxong:
I'm against server-side rendering logic because we'll need a webserver; plain HTML or client-side rendering are better options.
I think client-side rendering is better and will become better as the site grows.
Currently there are so many games that the page HTML takes the most time and size to load:
![image](https://user-images.githubusercontent.com/1083215/54498938-ea69c880-4960-11e9-879f-762bd389ea6b.png)
Most of that size is HTML tags. The raw data (YAML) is a fraction of that size.

Front end frameworks can also be big, but it's not as bad - they can be cached, and we could also pick a lightweight one like vue or preact: https://gist.github.com/Restuta/cda69e50a853aa64912d

piranha:
The thing is that client-side rendering will be never as fast as static HTML. 123 kb of html is a little! I wonder why it loads for 2 seconds, maybe it's a problem with web server? I'll look into it.

Disclosure: I'm working on JS apps since 2011, and for the last few years I'm at https://kasta.ua - you can check it out, it's pretty fast, but I'll be damned if OSGC isn't a magnitude faster! :)

killthekitten:
I'd also like to sort by number of :octocat: ⭐️ but that would require some db backend to effectively cache the numbers.

CommonLoon102:
Client side rendering is a very bad idea because it only works with JavaScript turned on. This site shouldn't require JS to check basic things, this site is basically a static list. Only web applications should require JS. Everything else should work without JS.

piranha:
@CommonLoon102 well there is no way you can sort something without JS (or adding server, but that sounds even more atrocious).

CommonLoon102:
[...]

PR #1126, opened 2020-08-26
Add basic sort function

pehsa:
#93 
I implemented a basic sorting function.

This will hide the main list, and sort the game entries by data-updated attribute. No duplicated list, the JS generate the sorted list on click.

The Filters are working on the sorted list.

The only problem I was unable to solve yet, is when the JS create the game list its search for all DD tag. This will get one game multiple times if it has two or more original game.
I tried to clear the data-parent and data-index attributes and find the only unique DDs, but not worked.

Any idea?


cxong:
Newly updated games are missing the "NEW" badge

cxong:
Oh ok I see why - this list is old, the newest item is from july

PR #1157, opened 2020-10-12
Separate pages for games

piranha:
So the idea is the following: Google really likes pages dedicated to a single topic. If we get those, it'll bring more traffic to site, in turn driving traffic to those games (and showing that there are a lot of them), which sounds like a good thing. 

Also the main page is really big right now, and Google complains about that. :)

This pull request contains a script (`render.py`) which seems to be doing everything we need from site generator - this allows to drop Cyrax as a dependency and be more flexible. It generates index.html and then a separate page for each game (or group of games). It's not very pretty right now - it's the same markup as for a game entry in a common list, and I think it could be better/bigger/etc:

image

After that we could put a link to those games and see for some time if it will change Google's results. In theory Google should start driving more traffic to an individual game page.

We could also generate pages for genres etc, there is some search traffic there, like that:

image

### DIFF
--- a/.cyrax.cfg
+++ b/.cyrax.cfg
@@ -1,2 +1,0 @@
-sitecallback: _ext.callback
-exclude: [games/*, originals/*, schema/*, templates/*, Makefile, README.md, requirements.txt, Dockerfile, vhost.conf, CHECKS, Pipfile*, LICENSE, dangerfile.js, package.json, yarn.lock, bunny-sync, node_modules/*]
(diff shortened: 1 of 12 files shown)


piranha:
Huh, it's finally deployed. Look here: https://5f843ca5d7723600071fdff1--osgameclones.netlify.app/ace-of-spades/

cxong (approved):

piranha:
Just in case there is a desire page of a single game could be significantly improved. :) Don't think I'll get to markup any time soon :(

PR #1161, opened 2020-10-19
Reworked the "Super Mario" section.

onpon4:
* "Super Mario" is not a game, it's a series of games. The games that are most often sources of inspiration are Super Mario Bros. and Super Mario World.
* The pucpr games are not open source.
* Mega Mario is a clone, not a remake.
* ReTux is not a clone, it's just loosely similar to Super Mario World. It's development is also no longer stagnant.
* Super Mario War is not a clone of or even similar to any Mario game, it's an entirely original game similar to (but heavily extended compared to) Jump'n'Bump.
* SuperTux is not a clone, it's just loosely similar to Super Mario World.
* Tux Builder is a clone of SuperTux, not of any Mario game.
* uMario is a clone, not a remake, of Super Mario Bros. (by osgameclones definitions).

### DIFF
--- a/games/h.yaml
+++ b/games/h.yaml
@@ -220,31 +220,31 @@
   url: http://hexgl.bkcore.com/
 
 - name: Hexoshi
-  type: clone
+  type: similar
   originals:
   - Super Metroid
-  repo: https://savannah.nongnu.org/projects/hexoshi/
-  url: http://hexoshi.nongnu.org/index.html
-  development: sporadic
+  repo: https://github.com/hexoshi/hexoshi
+  url: https://hexoshi.github.io
+  development: active
   status: playable
   lang:
   - Python
   license:
   - GPL3
   images:
-  - http://hexoshi.nongnu.org/screenshots/screenshot1.png
-  - http://hexoshi.nongnu.org/screenshots/screenshot2.png
-  - http://hexoshi.nongnu.org/screenshots/screenshot3.png
-  - http://hexoshi.nongnu.org/screenshots/screenshot4.png
-  - http://hexoshi.nongnu.org/screenshots/screenshot5.png
-  - http://hexoshi.nongnu.org/screenshots/screenshot6.png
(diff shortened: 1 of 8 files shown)


games/h.yaml
```diff
@@ -220,31 +220,31 @@
...
+  - https://hexoshi.github.io/screenshots/screenshot4.png
+  - https://hexoshi.github.io/screenshots/screenshot5.png
+  - https://hexoshi.github.io/screenshots/screenshot6.png
+  - https://hexoshi.github.io/screenshots/screenshot7.png
+  - https://hexoshi.github.io/screenshots/screenshot8.png
+  - https://hexoshi.github.io/screenshots/screenshot9.png
+  - https://hexoshi.github.io/screenshots/screenshot10.png
+  - https://hexoshi.github.io/screenshots/screenshot11.png
+  - https://hexoshi.github.io/screenshots/screenshot12.png
+  - https://hexoshi.github.io/screenshots/screenshot13.png
```
cxong, line 247:
```suggestion
  - https://hexoshi.github.io/screenshots/screenshot13.png
  updated: 2020-10-22
```

games/t.yaml
```diff
@@ -635,7 +635,7 @@
 
 - name: Tux Builder
   originals:
-  - Super Mario
+  - SuperTux
```
cxong, line 638:
```suggestion
  - Super Mario World
```
Since the game is extremely similar to SuperTux, which is in turn a game like Super Mario World, it's better to leave the original as SMW

cxong (changes requested):
Thanks for the updates. Some notes:

- Please leave the games without licenses in. We made a decision that osgameclones would include such games, but mark the license as `As-is`. It's not ideal since this may be misleading, so we can discuss enhancing the site in another issue. But we'd very much like to include such games in this site anyway
- For Super Mario War, since the gameplay is very similar to jump n bump, we can add said game as an original instead of removing it

PR #1186, opened 2020-11-22
Differentiate the C&C engines projects

Mailaender:
as @OpenRA wants to go beyond being a clone and is also not based on original source or binary code, but a clean room implementation. The others want to stay pure to the original intentionally.

### DIFF
--- a/games/c.yaml
+++ b/games/c.yaml
@@ -407,7 +407,7 @@
   originals:
   - 'Command & Conquer: Red Alert'
   repo: https://github.com/TheAssemblyArmada/Chronoshift
-  type: remake
+  type: clone
   status: playable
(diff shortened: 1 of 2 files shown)


cxong (changes requested):
This is using the wrong terminology; "remake" is used for games that replace the original assets or executable; see the home page at https://osgameclones.com/
```
A remake is a game where the executable and sometimes the assets as well are remade open source. Some of these games aren't exact remakes but evolution of original ones, which were eventually open sourced.
A clone is a game which is very similar to or heavily inspired by a game or series.
```

Mailaender:
Sorry, I misunderstood those terms.

PR #1199, opened 2020-12-08
Added re-private-eye

neuromancer:


cxong (changes requested):
Need to add the original game

neuromancer:
I think it should be fixed.

cxong (approved):

PR #1240, opened 2021-01-30
Fixed the GTA2.NET URL as it was referencing a copy.

Mailaender:
It even says on https://github.com/andrecarlucci/gta2net
> Automatically exported from code.google.com/p/gta2net

Also no activity on that fork either.

### DIFF
--- a/games/g.yaml
+++ b/games/g.yaml
@@ -453,8 +453,8 @@
   originals:
   - Grand Theft Auto 2
   type: remake
-  updated: 2018-10-05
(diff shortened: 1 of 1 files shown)


cxong (changes requested):
We don't want to link to google code repos because that service is dead

Mailaender:
There is no other URL as the game project is dead as well.

wonea:
Merging in as the repo it's pointing to is empty

PR #1264, opened 2021-03-08
Update freedroidclassic

cxong:


wonea (commented):
Point to website not download site

cxong:
Website doesn't mention freedroid classic however, it is only mentioned in the download page
The main website is devoted to the spinoff freedroid rpg

PR #1270, opened 2021-03-14
Added X-Settlers

Mailaender:
a native Settlers III clone.

### DIFF
--- a/games/x.yaml
+++ b/games/x.yaml
@@ -83,6 +83,24 @@
   updated: 2020-12-16
   url: http://xmoto.tuxfamily.org/
 
+- name: X-Settlers
+  development: active
+  images:
+  - https://raw.githubusercontent.com/Skarmux/X-Settlers/master/xsettlers/docs/screenshots/xsettlers_textured.png
+  lang:
+  - C++
+  framework:
+  - SDL2
+  license:
+  - Apache
+  content: commercial
+  originals:
+  - The Settlers III
+  repo: https://github.com/Skarmux/X-Settlers
+  type: remake
+  status: unplayable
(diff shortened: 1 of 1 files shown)


games/x.yaml
```diff
@@ -83,6 +83,21 @@
...
   url: http://xmoto.tuxfamily.org/
 
+- name: X-Settlers
+  development: active
+  images:
+  - https://raw.githubusercontent.com/Skarmux/X-Settlers/master/xsettlers/docs/screenshots/xsettlers_textured.png
+  lang:
+  - C++
+  license:
+  - Apache-2.0
```
caiotava, line 93: The `Apache-2.0` license doesn't exist in the validation schema: https://github.com/opengaming/osgameclones/blob/master/schema/games.yaml#L110

I think you could just use `Apache` in this case, what do you think?

Mailaender: Yes, fixed it.

caiotava (approved):

games/x.yaml
```diff
@@ -83,6 +83,22 @@
   updated: 2020-12-16
   url: http://xmoto.tuxfamily.org/
 
+- name: X-Settlers
+  development: active
+  images:
+  - https://raw.githubusercontent.com/Skarmux/X-Settlers/master/xsettlers/docs/screenshots/xsettlers_textured.png
+  lang:
+  - C++
```
cxong, line 91:
```suggestion
  - C++
  framework:
  - SDL2
```

Mailaender: Ah, nice spotting.

cxong (approved):

PR #1272, opened 2021-03-14
Fix docker build

caiotava:
pypa/pipenv#1050

Pipenv doesn't support multiple python versions and the base docker
image for nginx:1.17-alpine has python 3.8 and I want to enable
environments with python version greater than 3.7 to use pipenv.


cxong (approved):

piranha:
Hah, I was overly cautious :-)

PR #1277, opened 2021-03-21
Add Makefile recipes to help run the server with Docker

caiotava:


Makefile
```diff
@@ -1,5 +1,17 @@
...
 	pipenv run ./render.py
 
 prod:
 	./render.py && htmlmin _build/index.html _build/index.html
+
+docker-build:
+	docker build -t opengaming/osgameclones .
+
+docker-run:
+	-docker rm -f osgameclones-site
```
cxong, line 13: what does the leading `-` do here?

caiotava: It ignores the error returned by the `docker rm -f osgameclones-site` command.

I put it here because this command will try to remove the old docker container, before creating a new one.

If I run this command and the docker container doesn't exist, it'll return an error and abort the make recipe.
An example for this scenario is the first time that you run `make docker-run`

cxong (approved):

PR #1282, opened 2021-03-28
Fix Wikipedia link for Powerslave

jaens:
Wikipedia is case sensitive, the current link goes to an Iron Maiden album :metal:


wonea:
Thanks, good spot!

PR #1287, opened 2021-04-03
Added mc2

Mailaender:
another Mech Commander 2 clone based on the partial source code release from @Microsoft


alariq:
Thanks for adding. As for the status: the game is playable on both Linux and Windows. I've even passed it :-), well I had to check that it works. But there is no binaries, so I am not sure what to put in the status field :-/

PR #1295, opened 2021-04-14
Update prboom+

cxong:


caiotava:
It might be a good thing to add SDL2 as framework. What do you think?

games/p.yaml
```diff
@@ -626,9 +626,11 @@
   - Doom
   - Heretic
   - Hexen
```
cxong, line 628:
```suggestion
  - Hexen
  framework:
  - SDL2
```

caiotava (approved):

PR #1340, opened 2021-06-17
Update opentomb

cxong:


games/o.yaml
```diff
@@ -1845,17 +1845,24 @@
   - https://opentomb.github.io/img/media/03.jpg
   lang:
   - C++
+  framework:
+  - OpenGL
+  - SDL
```
caiotava, line 1850: I guess this project use SDL2 instead of SDL.

cxong, line 1850:
```suggestion
  - SDL2
```

PR #1342, opened 2021-06-18
Add Mineclone2

basxto:
Add Mineclone2, a Minetest game that aims to clone Minecraft

### DIFF
--- a/games/m.yaml
+++ b/games/m.yaml
@@ -379,7 +379,7 @@
    - https://mindustrygame.github.io/images/2.png
    - https://mindustrygame.github.io/images/3.png
 
-- name: Mineclone2
+- name: MineClone 2
   development: active
   images:
   - https://satoshiupload.com/images/DWimu2CTrE.png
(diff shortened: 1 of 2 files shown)


games/m.yaml
```diff
@@ -379,6 +379,31 @@
...
+  development: active
+  images:
+  - https://satoshiupload.com/images/DWimu2CTrE.png
+  - https://satoshiupload.com/images/6qS9a8vrCP.png
+  - https://i.imgur.com/1oOMc1L.png
+  - https://i.imgur.com/R7B0E54.png
+  lang:
+  - Lua
+  framework:
+  - Minetest Engine
```
cxong, line 392: you can add this to `dangerfile.js` to suppress the warning

cxong: Would `Minetest` be a better name for this engine?

cxong (approved):

games/m.yaml
basxto: It doesn't call itself that, but neither does Spring RTS and those two are very similar, since they view mods as the actual games.
Though there is also a [Minetest Game](https://content.minetest.net/packages/Minetest/minetest_game/) default game, which Spring RTS does not have to my knowledge.

PR #1343, opened 2021-06-21
Github actions

cxong:


cxong:
@piranha need your help here; I'm converting CI from travis to github actions, because I migrated to travis-ci.com but it unfortunately has limited credits for building. I think I'm almost done with the conversion, one thing left is the netlify auth token.

I think we need to add the auth token `NETLIFY_AUTH_TOKEN` as a repo secret here: https://github.com/opengaming/osgameclones/settings/secrets/actions
Documentation for the build step is here: https://github.com/marketplace/actions/netlify-actions

piranha:
@cxong done!

PR #1345, opened 2021-06-26
Update & fix Minetest and Minecraft

basxto:
Minecraft isn't a First Person Shooter, though it has a bow and such. It has elements of an Action RPG.

Minetest is only similar, the core always focused more on the sandboxing part and does not have the RPG elements, minecraft has. They additionally developed apart over the last 10 years.
Minetest had mobs at first, but removed them later. It in general transformed into an engine with a default game (Minetest Game), but that one still lacks mobs, NPCs, XP, enchanting, dimensions and such. It basically only has mining, crafting, building and farming. And without additional mods it's also not possible to build machines like it's possible with Minecraft's redstone.

> The gameplay is mostly sandbox-style

https://content.minetest.net/packages/Minetest/minetest_game/

### DIFF
--- a/originals/m.yaml
+++ b/originals/m.yaml
@@ -364,14 +364,6 @@
     - Arcade
 
 - name: Minecraft
-  names:
-  - 'Minecraft: Java Edition'
-  - 'Minecraft: Windows 10 Edition'
-  - 'Minecraft: Xbox 360 Edition'
-  - 'Minecraft: New Nintendo 3DS Edition'
-  - 'Minecraft: Wii U Edition'
-  - 'Minecraft: Playstation 3 Edition'
-  - 'Minecraft: Playstation Vita Edition'
   external:
     wikipedia: Minecraft
     website: https://www.minecraft.net  
--- a/originals/m.yaml
+++ b/originals/m.yaml
@@ -364,15 +364,40 @@
     - Arcade
 
 - name: Minecraft
+  names:
+  - 'Minecraft: Java Edition'
+  - 'Minecraft: Windows 10 Edition'
+  - 'Minecraft: Xbox 360 Edition'
+  - 'Minecraft: New Nintendo 3DS Edition'
+  - 'Minecraft: Wii U Edition'
+  - 'Minecraft: Playstation 3 Edition'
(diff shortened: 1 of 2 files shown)


originals/m.yaml
```diff
@@ -364,15 +364,40 @@
...
 
 - name: Minecraft
+  names:
+  - 'Minecraft: Java Edition'
+  - 'Minecraft: Windows 10 Edition'
+  - 'Minecraft: Xbox 360 Edition'
+  - 'Minecraft: New Nintendo 3DS Edition'
+  - 'Minecraft: Wii U Edition'
+  - 'Minecraft: Playstation 3 Edition'
+  - 'Minecraft: Playstation Vita Edition'
```
cxong, line 374: I don't think it's necessary to add these alternate names; the platform information is enough

basxto: I’m thinking about renaming it to
```
- 'Minecraft: Java Edition'
- 'Minecraft: Legacy Console Edition'
- 'Minecraft: Bedrock Edition'
- 'Minecraft: Pi Edition'
```

Which would make sense for the purpose of the site, since features and gameplay is not neccesarily the same. And they share some platforms.

So different projects could try to clone different editions.

Java Edition: (formerly just Minecraft)
- Windows
- MacOS
- Linux

Legacy Console Edition: (initial console ports by a third party)
- Xbox 360
- Xbox One
- PS3
- PS4
- PS Vita
- Wii U
- Nintendo Switch

Bedrock Edition: (C++ recreation of the engine)
- Android/Fire OS
- iOS
- Windows 10
- Xbox One
- Nintendo Switch
- PlayStation 4

Pi Edition is a free version for ARM only. With cut down functionality, but programming language APIs.

cxong: The names here are to help people find games that can appear under very different names, for example "Resident Evil 2"/"Biohazard". For minecraft this isn't necessary, because they all have the "minecraft" name in common.
If there are clones that only implement say the legacy console features, this can go in the `info` field.
Having too many names makes it appear bloated in the webpage.

cxong (approved):

PR #1351, opened 2021-07-01
Update ironseed

cxong:


mnalis:
FYI, dozens of screenshots for `ironseed_fpc` can be found at https://screenshots.debian.net/package/ironseed so feel free to add them

PR #1352, opened 2021-07-04
add OpenAWE, a remake of Alan Wake

kparal:


kparal:
The project builds fine locally.

originals/a.yaml
```diff
@@ -115,6 +115,22 @@
     genre:
     - Action
 
+- name: Alan Wake
```
cxong, line 118: looks like this and aladdin aren't sorted properly (should be above albion)
could you please move them up? thanks

kparal: Sure, fixed. Please note that `games/o.yaml` is also wrongly sorted, because humans are often unsure about the sort order of letters, spaces and dashes (myself included). Perhaps a sorting script would be a good idea ;-)

kparal:
Is anything else required? I'm happy to do more changes, if needed.

originals/a.yaml
```diff
@@ -82,6 +82,29 @@
     theme:
     - Fantasy
 
+- name: Aladdin
+  external:
+    wikipedia: /Disney's_Aladdin_(Virgin_Games_video_game)
```
cxong, line 87:
```suggestion
    wikipedia: Disney's Aladdin (Virgin Games video game)
```

originals/a.yaml
```diff
@@ -82,6 +82,29 @@
...
+  meta:
+    genre:
+    - Action
```
cxong, line 90:
```suggestion
    - Platform
```

PR #1360, opened 2021-07-17
add Ambermoon.net, a remake of Ambermoon

kparal:
Please note, while I used `content: free`, because the content is bundled in the release archive and free for the user, I believe the author actually uses the original assets from Ambermoon, i.e. most probably copyrighted content. I assume they are relying on its abandonware status and take the liberty to distribute it. I'm not sure if the `content` field should be changed to `commercial` or not.

### DIFF
--- a/games/a.yaml
+++ b/games/a.yaml
@@ -225,7 +225,7 @@
   type: remake
   updated: 2014-01-12
   url: http://retrospec.sgn.net/users/ignacio/a8i.htm
-  
+
 - name: Alive_Reversing
   type: remake
   originals:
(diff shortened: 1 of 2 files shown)


games/a.yaml
```diff
@@ -255,6 +255,28 @@
...
+  originals:
+    - Ambermoon
+  repo: https://github.com/Pyrdacor/Ambermoon.net
+  development: very active
+  status: playable
+  lang:
+    - 'C#'
+  license:
+    - GPL3
+  content: free
```
cxong, line 269:
```suggestion
  content: commercial
```
since abandonware isn't a license

games/a.yaml
```diff
@@ -255,6 +255,28 @@
...
 
+- name: Ambermoon.net
+  type: remake
+  originals:
+    - Ambermoon
+  repo: https://github.com/Pyrdacor/Ambermoon.net
+  development: very active
+  status: playable
+  lang:
+    - 'C#'
```
cxong, line 266:
```suggestion
    - C#
    framework:
    - .NET
```

cxong (changes requested):

kparal:
Updated per review.

cxong (approved):

kparal:
What is the next step now?

PR #1502, opened 2021-10-25
Add siboot

zetadeceleste:


games/s.yaml
```diff
@@ -401,6 +401,22 @@
   type: remake
   updated: 2018-06-15
 
+  name: 'Trust & Betrayal: The Legacy of Siboot'
```
cxong, line 404:
```suggestion
- name: 'Trust & Betrayal: The Legacy of Siboot'
```

cxong (changes requested):

games/s.yaml
zetadeceleste: I'm sorry, I don't get the change.

cxong: this is yaml syntax for creating a new array item; without it this would be part of the previous item

zetadeceleste: Ok! gotcha!

games/s.yaml
```diff
@@ -401,6 +401,22 @@
...
+    website: 'https://siboot.org/'
+  wikipedia: 'https://en.wikipedia.org/wiki/Trust_%26_Betrayal:_The_Legacy_of_Siboot'
+  platform:
+  - DOS
+  meta:
+    genre:
+    - Educational
+    - Simulation
+    - Strategy
+  theme:
```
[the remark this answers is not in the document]
zetadeceleste: Now it's ok?

games/s.yaml
```diff
@@ -401,6 +401,22 @@
   type: remake
   updated: 2018-06-15
 
+- name: 'Trust & Betrayal: The Legacy of Siboot'
```
cxong, line 404: this looks like the wrong file

cxong (changes requested):

games/s.yaml
zetadeceleste: I added at originals too, but i don't know if it's ok

cxong: the files under games/ are for clones
the originals go under originals/
the two types of files have different schemas, so this stuff doesn't belong here

cxong (changes requested):

games/s.yaml
zetadeceleste: I checked this syntax with https://osgameclones.com/add_original.html as you recommend and pushed,
would you tell me please if it's correct?

zetadeceleste: So, you are saying instead of adding this game at /games I should move it to /originals?
It's that correct?

cxong: correct; please refer to the schema and readme

cxong (changes requested):

games/s.yaml
zetadeceleste: done, please let me know! thanks

cxong (approved):

Issue #1492, opened 2021-10-23
Add script for updating updated field in game's description


AlexHladin:
A lot of games' source code is stored on Github/Gitlab/Bitbucket and all of them have the field `updated` that refers to the date of the last commit in the repo. 
It will be a good point to create a javascript/python script that could grab the last update date of all games.

For example, `Bedlam 2: Absolute Bedlam` has `updated` field that points to 2021-10-03
![image](https://user-images.githubusercontent.com/7945837/138566049-38482a37-6043-4af1-8613-bda0c4ccc775.png)

but the last update has been 2021-10-26.
It is obvious that this information should be fetched automatically by the script or any other tool.

I hope the description is clear. On top of that, I would like to create the script that I describe.

PR #1507, opened 2021-10-31
Add script for updating games statuses

AlexHladin:
The PR adds functionality for updating games statuses as described in the propose #1492

### DIFF
--- a/scripts/update-development-status.py
+++ b/scripts/update-development-status.py
@@ -1,0 +1,112 @@
+"""
+This script updates games development status
+
+To run, install from pip:
+- pygithub
+- python-gitlab
+
+Add environment variables:
+- GH_TOKEN (see https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#creating-a-token)
+- GL_TOKEN (see https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#create-a-personal-access-token)
+"""
+import os
+import re
+import yaml
+
+from pathlib import Path
+from datetime import datetime, timedelta
+from github import Github, GithubException
+from gitlab import Gitlab
+
+GH_REGEX = re.compile(r'https://github.com/([^/]+)/([^/]+)')
+GL_REGEX = re.compile(r'https://gitlab.com/([^/]+/[^/]+)')
+
+
+def main():
+    gh = Github(os.environ["GH_TOKEN"])
+    gl = Gitlab('https://gitlab.com', private_token=os.environ["GL_TOKEN"])
+
+    for filename in Path('games').iterdir():
+        if not filename.is_file() or filename.suffix != '.yaml':
+            continue
+
+        games = yaml.safe_load(open(filename, encoding='utf-8'))
+        for game in games:
+            repo_url = game.get('repo', '')
+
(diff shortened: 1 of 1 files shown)


scripts/update-development-status.py
```diff
@@ -0,0 +1,112 @@
...
+    gh = Github(os.environ["GH_TOKEN"])
+    gl = Gitlab('https://gitlab.com', private_token=os.environ["GL_TOKEN"])
+
+    for filename in Path('games').iterdir():
+        if not filename.is_file() or filename.suffix != '.yaml':
+            continue
+
+        games = yaml.safe_load(open(filename, encoding='utf-8'))
+        for game in games:
+            repo_url = game.get('repo', '')
```
cxong, line 35: should continue if there is no repo, because getting latest commit date is impossible

scripts/update-development-status.py
```diff
@@ -0,0 +1,112 @@
...
+                game['development'] = 'sporadic'
+            else:
+                game['development'] = 'halted'
+
+        yaml.dump(games, open(filename, 'w', encoding='utf-8'), sort_keys=False)
+        print(filename, 'has been updated')
+
+
+def is_github_repo(repo):
+    return 'https://github.' in repo
```
cxong, line 61: nit: repo.startswith(...)

scripts/update-development-status.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def is_gitlab_repo(repo):
+    return 'https://gitlab.' in repo
```
cxong, line 65: should check that the full host is gitlab.com
Some repos use a self-hosted gitlab; the GL_TOKEN isn't going to work there

scripts/update-development-status.py
```diff
@@ -0,0 +1,112 @@
...
+
+    project_namespace = match.groups()[0]
+    project = gl.projects.get(project_namespace)
+
+    commits = project.commits.list()
+
+    return datetime.strptime(
+        ''.join(commits[0].committed_date.rsplit(':', 1)),
+        '%Y-%m-%dT%H:%M:%S.%f%z'
+    ).replace(tzinfo=None)
```
cxong, line 108: question: why remove tzinfo? What timezone does gitlab dates return as?

cxong: if it makes it more convenient, we can work with dates instead of datetime

cxong (changes requested):

scripts/update-development-status.py
AlexHladin: GitLab returns commit date in user local date-time, not by UTC :(

AlexHladin: On top of that, it is not convenient to work with date, date arithmetic works with date-time only

scripts/update-development-status.py
AlexHladin: Updated

scripts/update-development-status.py
AlexHladin: Fixed

scripts/update-development-status.py
AlexHladin: Fixed

cxong (approved):

PR #1510, opened 2021-11-02
Update freerails

cxong:


cxong:
cc @Trilarion

cxong:
Thanks for the info; will change to `unplayable`

Trilarion:
I think it's good now.

PR #1535, opened 2021-12-02
switch to poetry

piranha:
hopefull netlify will generate deploy preview :)


piranha:
Woooooooo it works, wow. :)

PR #1556, opened 2021-12-22
Add max payne

cxong:


isage:
Technically, that isn't port. It's a bundled wrapper to run android version.
Following that logic you can add any game, just because wine can run it on linux.
Game doesn't magically become opensource with that.

fgsfdsfgs:
Nor is it a remake or a clone.

PR #1629, opened 2022-02-27
Get edit form data from JSON get (fixes #1595)

### LINKED ISSUE #1595: Optimize edit form builds
Most of the build time is taken up by the specific edit game / edit clone forms. This can be optimized by only rendering one set of add/edit game/clone forms, but having an optional javascript load step which loads form values from the yaml files.

cxong:
Use the same add original/game forms, but with an AJAX get to load game data.

### DIFF
--- a/render.py
+++ b/render.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import html
+import json
 import os, os.path as op
 import shutil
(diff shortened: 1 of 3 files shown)


cxong:
Speeds up build-and-deploy from 2m 24s to 52s

PR #1669, opened 2022-04-02
Update KeeperFX, remove KeeperFX Unofficial

hansejo:
Progress has restarted under the original project. This change reflects this

### DIFF
--- a/games/k.yaml
+++ b/games/k.yaml
@@ -92,9 +92,11 @@
   originals:
   - Dungeon Keeper
   repo: https://github.com/dkfans/keeperfx
-  type: remake
+  type: tool
   updated: 2022-04-02
   url: https://lubiki.keeperklan.com/html/dk_keeperfx.php
+  video:
+    youtube: JwDdLynZPag
   
 - name: KevEdit
   type: tool
--- a/games/k.yaml
+++ b/games/k.yaml
@@ -88,27 +88,14 @@
   - C
   license:
   - As-is
-  development: halted
+  development: active
   originals:
   - Dungeon Keeper
   repo: https://github.com/dkfans/keeperfx
   type: remake
-  updated: 2020-05-18
+  updated: 2022-04-02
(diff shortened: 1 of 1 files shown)


games/k.yaml
```diff
@@ -88,27 +88,14 @@
...
   license:
   - As-is
-  development: halted
+  development: active
   originals:
   - Dungeon Keeper
   repo: https://github.com/dkfans/keeperfx
   type: remake
-  updated: 2020-05-18
+  updated: 2020-08-25
```
cxong, line 96:
```suggestion
  updated: 2022-04-03
```

cxong (commented):
also it would be good to add some images/video; check the netlify deploy for what it will look like

games/k.yaml
hansejo: Whoops, thought updated meant something different thanks!

hansejo:
@cxong  I would like to add a screenshot, but I haven't built it yet. And no other screenshots I can find.

hansejo:
@cxong Maybe we shouldn't consider this an engine remake. Main website says that this converts the original DK into a DLL which it then loads. So this is more akin to a patch. Thoughts on this?

games/k.yaml
```diff
@@ -88,27 +88,14 @@
   - C
   license:
   - As-is
-  development: halted
+  development: active
   originals:
   - Dungeon Keeper
   repo: https://github.com/dkfans/keeperfx
   type: remake
```
cxong, line 95:
```suggestion
  type: tool
```

games/k.yaml
```diff
@@ -88,27 +88,14 @@
...
-  updated: 2020-05-18
+  updated: 2022-04-02
   url: https://lubiki.keeperklan.com/html/dk_keeperfx.php
```
cxong, line 97:
```suggestion
  url: https://lubiki.keeperklan.com/html/dk_keeperfx.php
  video:
    youtube: JwDdLynZPag
```

PR #1696, opened 2022-05-03
X-COM updates

hansejo:
Firstly: remove dupe game UFO: Enemy Unknown (It's just another name for X-COM: UFO Defense). Already added as an alt-name under original X-COM entry.

Then, a few updates to X-COM-related games.

### DIFF
--- a/originals/u.yaml
+++ b/originals/u.yaml
@@ -1,15 +1,3 @@
-- name: 'UFO: Enemy Unknown'
-  external:
-    wikipedia: 'UFO: Enemy Unknown'
-  meta:
-    genre:
-    - Strategy
-    - Real-Time Strategy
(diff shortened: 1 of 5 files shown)


originals/u.yaml
```diff
@@ -1,15 +1,3 @@
-- name: 'UFO: Enemy Unknown'
```
cxong, line 1: this can be an alternate name for XCOM UFO Defense (add under the `names` list)

cxong (approved):

originals/u.yaml
hansejo: This is already [added](https://github.com/opengaming/osgameclones/blob/19297f299255ca4de4f34184925407418d82fdb6/originals/x.yaml#L27)

PR #1699, opened 2022-05-05
Add Fish Fillets NG

nextghost:
Hello, the game list is missing Fish Fillets NG, an old Czech puzzle game commercially released in 1997 and open-sourced in 2004.

### DIFF
--- a/games/f.yaml
+++ b/games/f.yaml
@@ -176,6 +176,29 @@
   url: http://www.newbreedsoftware.com/fop/
   info: A Dandy clone with Gauntlet influences. Choose from 8 classes and roam the maze taking treasure, vanquishing monters, and finding the exit. Use explosives against enemies and keys on locked doors.
 
+- name: Fish Fillets NG
+  originals:
+  - Fish Fillets
+  type: remake
+  repo: 'https://sourceforge.net/projects/fillets/'
+  url: 'http://fillets.sourceforge.net/'
+  development: complete
+  status: playable
+  content: open
+  lang:
+  - C++
+  - Lua
+  framework:
+  - SDL
+  license:
+  - GPL2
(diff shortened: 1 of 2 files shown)


games/f.yaml
```diff
@@ -176,6 +176,30 @@
...
 
+- name: Fish Fillets NG
+  originals:
+  - Fish Fillets
+  type: remake
+  repo: 'https://sourceforge.net/projects/fillets/'
+  url: 'http://fillets.sourceforge.net/'
+  development: complete
+  status: playable
+  content: free
```
cxong, line 187:
```suggestion
  content: open
```
I read that the content is also GPL; note OSGC uses "free" as in "beer"

games/f.yaml
```diff
@@ -176,6 +176,30 @@
...
+  - Lua
+  framework:
+  - SDL
+  license:
+  - GPL2
+  info: Cross-platform port of the original game sources.
+  updated: 2022-05-05
+  images:
+  - 'http://fillets.sourceforge.net/img/screenshot/ffng-pyramid.png'
+  - 'http://fillets.sourceforge.net/img/screenshot/ffng-chest.png'
```
cxong, line 199: images should be HTTPS otherwise they don't show when embedded

cxong (changes requested):

games/f.yaml
nextghost: Thanks, fixed.

games/f.yaml
nextghost: Unfortunately, the SourceForge project site does not support HTTPS so I hade to replace both images with one from Wikipedia.

cxong (approved):

PR #1708, opened 2022-05-08
added doom to games list

nather22:


games/d.yaml
```diff
@@ -617,6 +617,26 @@
   content: commercial
   updated: 2019-08-06
 
+- name: Doom
+  lang:
+  - C
+  development: complete
+  originals:
```
cxong, line 624:
```suggestion
  original:
```

cxong (changes requested):
Need to add license

PR #1711, opened 2022-05-09
add doom

nather22:


games/d.yaml
```diff
@@ -617,6 +617,28 @@
   content: commercial
   updated: 2019-08-06
 
+- name: Doom
+  lang:
+  - C
+  license:
+  - GPL1
```
cxong, line 624:
```suggestion
  - Custom
```
This release only has https://github.com/id-Software/DOOM/blob/master/linuxdoom-1.10/DOOMLIC.TXT - I can't find the GPL license in the repo

cxong (changes requested):

games/d.yaml
nather22: Sorry, you're right. This release is under their own souce license (DSL).

cxong (approved):

PR #1739, opened 2022-06-06
Added Xenoveritas' Abuse

Mailaender:
Another fork of a fork of the same source port, but with a modern built system, which is why I used it for https://github.com/flathub/flathub/pull/3208.

### DIFF
--- a/games/a.yaml
+++ b/games/a.yaml
@@ -20,6 +20,27 @@
   url: http://abuse.zoy.org/
   content: swappable
 
+- name: Abuse
+  originals:
+  - Abuse
+  type: remake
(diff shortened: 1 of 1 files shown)


games/a.yaml
```diff
@@ -20,6 +20,27 @@
   url: http://abuse.zoy.org/
   content: swappable
 
+- name: Abuse
+  originals:
+  - Abuse
+  type: official
```
cxong, line 26:
```suggestion
  type: remake
```

cxong (changes requested):
this is a fork from the official source code release so it shouldn't have type official

cxong (approved):

PR #1824, opened 2022-08-28
add the OpenGOAL project

romatthe:
Since other decompilation-related projects are now included in the list as well (`sm64ex`, `oot`, etc.), I believe OpenGOAL absolutely deserves a spot on this list!

### DIFF
--- a/games/o.yaml
+++ b/games/o.yaml
@@ -218,6 +218,33 @@
   video:
     youtube: vnUIWB7Mixw
 
+- name: OpenGOAL
+  originals: 
+  - 'Jak and Daxter: The Precursor Legacy'
+  type: remake
+  repo: https://github.com/open-goal/jak-project
+  url: https://opengoal.dev/
+  development: active
+  status: playable
+  content: commercial
+  langs:
+  - C++
+  - GOAL
+  licenses:
+  - ISC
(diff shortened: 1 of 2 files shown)


romatthe:
Since github/linguist doesn't define `GOAL`, I've removed it from the languages in which the project is written, though I'd still stay it's somewhat/semi correct.

We can leave it as is, or add a generic `Lisp` tag, I don't know which one is more appropriate.

games/o.yaml
```diff
@@ -218,6 +218,32 @@
...
+  originals: 
+  - 'Jak and Daxter: The Precursor Legacy'
+  type: remake
+  repo: https://github.com/open-goal/jak-project
+  url: https://opengoal.dev/
+  development: active
+  status: playable
+  content: commercial
+  langs:
+  - C++
```
cxong, line 231:
```suggestion
  - C++
  - Common Lisp
```

cxong (approved):

cxong:
if you want, you can raise an issue / PR to add GOAL to linguist, if it is worth it

romatthe:
No, it's definitely not worth it to do that. It's fine as it is I believe, thanks for suggesting though.

PR #1830, opened 2022-09-01
Update FreeCS and add FreeHL

Unargent:


games/f.yaml
```diff
@@ -597,28 +597,50 @@
...
+  - QuakeC
+  licenses:
+  - ISC
+  development: active
+  originals:
+  - Half-Life
+  status: semi-playable
+  repo: https://github.com/eukara/freehl
+  type: remake
+  content: swappable
```
cxong, line 637:
```suggestion
  content: commercial
```

cxong (changes requested):

games/f.yaml
Unargent: I changed the content type to commercial. But maybe you should add the ability to specify separately content for single-player and multiplayer? Because the full version of the singleplayer game needs commercial content, but for multiplayer enough free (shareware) content.

cxong (approved):

games/f.yaml
cxong: this seems like a rare use case. If it's helpful we can add this in the free-text `info` field

PR #1874, opened 2022-10-17
2nd attempt

cricri-pingouin:
Online form does seem to create the proper yaml structure?
Changed tabulation manually and trying again.


cricri-pingouin:
Ok, now found out I have some - in my strings, need quotes! So manual editing has drawbacks!

PR #1883, opened 2022-10-17
Update s.yaml

cricri-pingouin:
Added Same Game:
https://en.wikipedia.org/wiki/SameGame

### DIFF
--- a/originals/s.yaml
+++ b/originals/s.yaml
@@ -19,6 +19,13 @@
     - Post-Apocalyptic
     - Sci-Fi
 
+- name: Same Game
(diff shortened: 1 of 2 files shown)


originals/s.yaml
```diff
@@ -19,6 +19,13 @@
     - Post-Apocalyptic
     - Sci-Fi
 
+- name: Same Game
```
cxong, line 22:
```suggestion
- name: SameGame
```

cxong (changes requested):

PR #1885, opened 2022-10-18
Trying adding games again

cricri-pingouin:
Also changed original "Same Game" to "SameGame", as suggested.

### DIFF
--- a/originals/s.yaml
+++ b/originals/s.yaml
@@ -19,6 +19,13 @@
     - Post-Apocalyptic
     - Sci-Fi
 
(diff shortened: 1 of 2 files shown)


originals/s.yaml
```diff
@@ -19,6 +19,13 @@
     - Post-Apocalyptic
     - Sci-Fi
 
+- name: Same Game
```
cxong, line 22:
```suggestion
- name: SameGame
```

cxong (changes requested):

PR #1889, opened 2022-10-21
Add OceanPop - A HunniePop Clone

naddoska:
Hello Osgameclones Team! 

I added @sharkwouter s HunniePop Clone called 'OceanPop' to your List of Clones! 
Hope I did everything right. 
Have a good day :)

### DIFF
--- a/games/o.yaml
+++ b/games/o.yaml
@@ -13,6 +13,22 @@
   type: remake
   updated: 2016-02-08
 
+- name: OceanPop
+  images:
+  - https://raw.githubusercontent.com/sharkwouter/oceanpop/1.0/screenshot.jpg
+  langs:
+  - C++
+  licenses:
+  - MIT
+  development: sporadic
+  originals:
+  - HunniePop
+  info: HunniePop without the Hunnies
+  repo: https://github.com/sharkwouter/oceanpop
+  status: playable
+  type: clone
+  updated: 2022-03-09
+
 - name: Odamex
   langs:
   - C
--- a/games/o.yaml
+++ b/games/o.yaml
@@ -22,8 +22,8 @@
   - MIT
   development: sporadic
   originals:
-  - HunniePop
-  info: HunniePop without the Hunnies
+  - HuniePop
+  info: HuniePop without the Hunnies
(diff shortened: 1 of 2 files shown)


games/o.yaml
```diff
@@ -13,6 +13,22 @@
...
+  licenses:
+  - MIT
+  development: sporadic
+  originals:
+  - HuniePop
+  info: HuniePop without the Hunnies
+  repo: https://github.com/sharkwouter/oceanpop
+  status: playable
+  type: clone
+  updated: 2022-03-09
```
cxong, line 30:
```suggestion
  updated: 2022-10-23
```

games/o.yaml
```diff
@@ -13,6 +13,22 @@
...
+- name: OceanPop
+  images:
+  - https://raw.githubusercontent.com/sharkwouter/oceanpop/1.0/screenshot.jpg
+  langs:
+  - C++
+  licenses:
+  - MIT
+  development: sporadic
+  originals:
+  - HuniePop
```
cxong, line 25: I think something like bejeweled would be better here; I don't see any similarity with huniepop specifically except for the name "pop"
wondering what @sharkwouter thinks

games/o.yaml
```diff
@@ -13,6 +13,22 @@
...
   updated: 2016-02-08
 
+- name: OceanPop
+  images:
+  - https://raw.githubusercontent.com/sharkwouter/oceanpop/1.0/screenshot.jpg
+  langs:
+  - C++
+  licenses:
+  - MIT
+  development: sporadic
```
cxong, line 23:
```suggestion
  development: active
```

cxong (changes requested):

games/o.yaml
sharkwouter: Sure, I can eloborate. In Huniepop you have a specific amount of moves in which you have to reach a target score and you are able to move gems as far as you'd like. OceanPop implements both of those mechanics. This makes it quite different from Bejeweled.

naddoska:
@cxong I updated the requested changes :)

cxong (approved):

PR #1948, opened 2022-12-08
Update status for s games

cxong:


games/s.yaml
```diff
@@ -1762,6 +1811,20 @@
   updated: 2014-04-16
   url: http://methane.sourceforge.net/
 
+- name: Super Methane Brothers for Wii and GameCube
```
carstene1ns, line 1814: Maybe should rebrand this as "for Nintendo homebrew" since I released this for Switch and (at least planned to release) for 3DS.

PR #2089, opened 2023-04-16
Add Hishigata (clone) and Project Rhombus (original)

crse:


games/h.yaml
```diff
@@ -344,6 +344,23 @@
...
+  - Project Rhombus
+  type: clone
+  repo: 'https://github.com/LumpBloom7/hishigata'
+  development: sporadic
+  status: playable
+  content: open
+  langs:
+  - 'C#'
+  frameworks:
+  - osu!framework
```
cxong, line 358:
```suggestion
  - osu!
```

cxong: also add this value to frameworks in dangerfile.js to silence the warning

cxong (changes requested):

games/h.yaml
crse: I believe we should match the official name instead: `osu!framework` -> https://github.com/ppy/osu-framework

crse:
Should be good right now :)

PR #2115, opened 2023-05-14
Add beneath a steel sky #1818

cxong:


cxong:
@piranha is there a way to get deploy previews working in PRs?

piranha:
Argh I did not think about that :(

They have something about previews in readme, but it's marked _NOTE: This feature is only in alpha currently and is not available to the public!_. :(

There are other (non-official) actions to deploy pull requests as subdirs, I'll investigate that.

PR #2145, opened 2023-06-10
Add restunts (fixes #2118)

### LINKED ISSUE #2118: Add restunts
https://github.com/xor2003/restunts

cxong:


xor2003:
Hey probably you picked a wrong guy. What I have done is just translated game binary to source with fake-assembler which still require dosbox as library. Sure it can now run on 32/64 bit systems. Same as dosbox. Little faster. Also it is hard to compile.
There are other guys which are doing the decompilation. I don't finished it yet. Don't have time.
Maybe some day I will be able automaticaly decompile any game with my tools

PR #2164, opened 2023-06-23
Pushover was renamed to Domino-Chain

denilsonsa:


games/p.yaml
```diff
@@ -949,19 +949,19 @@
   type: remake
   updated: 2015-06-01
 
-- name: Pushover
+- name: Domino-Chain
```
cxong, line 952: please move this game to `d.yaml`

cxong (changes requested):

games/p.yaml
denilsonsa: Of course! 🤦
That's what happens when doing stuff in the middle of the night.
I've pushed a fix.

cxong (approved):

PR #2173, opened 2023-06-30
Add ENTERGRAM visual novels (Higurashi/Umineko) and their remakes

andOlga:


DCNick3:
Higurashi is not playable in shin right now due to differences in the VM opcodes. Not sure if this is the best way to represent it in the metadata...

Also, the license is not "as-is", but MPL (I probably need to add a LICENSE file :sweat_smile:)

andOlga:
Fixed up the license and removed Higurashi from the originals list for now. Can add it back later when it does become playable.

originals/h.yaml
```diff
@@ -222,6 +222,24 @@
     genres:
     - Real-Time Strategy
 
+- name: 'Higurashi: When They Cry Hou'
```
cxong, line 225:
```suggestion
- name: 'Higurashi: When They Cry'
```
I think it's better to refer to the original game here https://en.wikipedia.org/wiki/Higurashi_When_They_Cry

originals/h.yaml
```diff
@@ -222,6 +222,24 @@
...
+  external:
+    website: 'https://entergram.co.jp/higurashihou/'
```
cxong, line 227:
```suggestion
    website: 'https://entergram.co.jp/higurashihou/'
    wikipedia: Higurashi When They Cry
```
The wikipedia link is preferred as it is in english

originals/h.yaml
andOlga: The clones added here are specifically for the console remake. The original games are already source-available, though they are missing all of the extra features the remake offers, such as voice acting, visual improvements, etc.

originals/u.yaml
```diff
@@ -115,6 +115,24 @@
   - FM Towns
   - PC-98
 
+- name: 'Umineko: When They Cry Saku'
+  external:
+    website: 'https://entergram.co.jp/umineko/'
```
cxong, line 120:
```suggestion
    website: 'https://entergram.co.jp/umineko/'
    wikipedia: Umineko When They Cry
```

PR #2244, opened 2023-08-27
Add SuperTux Classic

basxto:
SuperTux Classic is the successor of TuxBuilder

Same framework, same author

### DIFF
--- a/games/s.yaml
+++ b/games/s.yaml
@@ -2587,6 +2587,27 @@
   updated: 2017-10-13
   url: https://supertux.org/
 
+- name: SuperTux Classic
+  development: active
+  frameworks:
+  - Godot
+  images:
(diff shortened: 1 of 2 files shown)


cxong (approved):

Alzter:
Let's go!! This is awesome.

Yes, Tux Builder was my first try making a SuperTux remake in the Godot game engine, but I was very inexperienced then, so the code quality was generally very poor.

SuperTux Classic is my second go at a Godot SuperTux remake. I made it after learning a lot more about Godot and coding in general from my Computer Science course. SuperTux Classic is complete, unlike Tux Builder, and has a full set of levels, worlds, customisation features, and even a better level editor than Tux Builder. Unlike Tux Builder, though, SuperTux Classic aims to recreate version 0.1.3 of SuperTux specifically (with some new features thrown in for good measure), rather than the latest version.

PR #2246, opened 2023-08-28
Add OpenVic

basxto:
It’s in an very early stage.

### DIFF
--- a/originals/v.yaml
+++ b/originals/v.yaml
@@ -30,6 +30,22 @@
   - PlayStation
   - Windows
 
+- name: Victoria II
+  names:
+  - victoria 2
+  external:
+    website: 'https://www.paradoxinteractive.com/games/victoria-ii/about'
+    wikipedia: 'https://en.wikipedia.org/wiki/Victoria_II'
+  platforms:
+  - Classic Mac OS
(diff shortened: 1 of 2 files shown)


basxto:
It would make sense to also add `Grand Strategy` for these kinds of games, but I’m not sure where that would belong.

originals/v.yaml
```diff
@@ -30,6 +30,22 @@
   - PlayStation
   - Windows
 
+- name: Victoria II
+  names:
+  - victoria 2
+  external:
+    website: 'https://www.paradoxinteractive.com/games/victoria-ii/about'
+    wikipedia: 'https://en.wikipedia.org/wiki/Victoria_II'
```
cxong, line 38:
```suggestion
    wikipedia: Victoria II
```

originals/v.yaml
```diff
@@ -30,6 +30,22 @@
   - PlayStation
   - Windows
 
+- name: Victoria II
+  names:
+  - victoria 2
```
cxong, line 35:
```suggestion
  - Victoria 2
```

cxong (changes requested):

PR #2316, opened 2023-10-30
Fix light theme preference is ignored

### LINKED ISSUE #1833: Website forgets light mode preference
If the user's machine is in dark mode but has manually specified the light theme, on page refresh the page still switches to dark theme

alexpalade:
## The problem cause
When the user clicks the "light mode" button, it sets "startInDarkTheme" to "false" in localStorage, but the init code doesn't test for "false" to toggle the light mode. This becomes apparent when "prefers-color-scheme" is "dark" and the website goes into dark mode instead.

## The fix
I check for _both_ the "dark" and "light" theme preferences in the localStorage first. If those aren't found, I set the theme based on the "prefers-color-scheme" media feature.

I wrote playwright [spec file](https://gist.github.com/alexpalade/ec6a86e6d49a232166194a5345fd3415) (tests at the bottom) to test multiple use cases. I also made [this state chart](https://github.com/opengaming/osgameclones/assets/1126946/dde9e6a5-acc6-4d91-a859-c4542f6f42f0) to clarify the problem.

An observation: the init code should always call either `setDarkTheme` or `setLightTheme` once to hide one of the mode toggle button icons. If none is called, both the moon and sun icons appear.

## Theme key renamed to "theme"

I renamed the localStorage theme key from "startInDarkTheme" (=true/false) to "theme" (=dark/light). I think it's more accurate to store the light theme preference as "theme=light" than "startInDarkTheme=false" which is misleading (and possibly the cause for this bug). It also allows to add an "auto" value (and a drop-down menu) in the future. If you dislike this change, I can revert it.

## Unrelated CSS color fix

I also snuck in a minor CSS typo fix (#**8**f6100) that was generating an error in the browser console. 🥷 I figured out the correct color from the surrounding context.

Closes #1833

### DIFF
--- a/static/dark-theme.js
+++ b/static/dark-theme.js
@@ -1,6 +1,6 @@
 const DARK_THEME_CLASS = 'dark-theme';
 const LIGHT_THEME_CLASS = 'light-theme';
-const DARK_THEME_KEY = 'startInDarkTheme';
+const THEME_KEY = 'theme';
 
 function setDarkTheme() {
   document.body.classList.remove(LIGHT_THEME_CLASS);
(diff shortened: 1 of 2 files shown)


cxong (approved):

cxong:
very nice!

PR #2359, opened 2023-12-04
Update x.yaml

Phredreeke:
Added note to XL Engine regarding supported games, and reference to it successor project The Force Engine

### DIFF
--- a/games/x.yaml
+++ b/games/x.yaml
@@ -232,6 +232,8 @@
   - OpenGL
   licenses:
   - As-is
+  info: >-
+  Only Dark Forces and Daggerfall were supported by public builds. This port has been superseded by The Force Engine by the same developer
   development: halted
   originals:
   - Dark Forces
--- a/games/x.yaml
+++ b/games/x.yaml
@@ -233,7 +233,7 @@
   licenses:
   - As-is
   info: >-
-  Only Dark Forces and Daggerfall were supported by public builds. This port has been superseded by The Force Engine by the same developer
+    Only Dark Forces and Daggerfall were supported by public builds. This port has been superseded by The Force Engine by the same developer
   development: halted
   originals:
   - Dark Forces
--- a/games/x.yaml
+++ b/games/x.yaml
@@ -233,7 +233,7 @@
   licenses:
   - As-is
   info: >-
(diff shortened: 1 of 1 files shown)


games/x.yaml
```diff
@@ -232,6 +232,8 @@
   - OpenGL
   licenses:
   - As-is
+  info: >-
+    Only Dark Forces and Daggerfall were supported by public builds. This port has been superseded by The Force Engine by the same developer
```
cxong, line 236: has this project been superseded? I could not find info on this in the XL engine github

Phredreeke: Yes. [Original announcement](https://theforceengine.github.io/2020/05/05/firstpost.html) [XL Engine github, stating the project is dead](https://github.com/luciusDXL/XL-Engine)

Phredreeke: LuciusDXL was the original developer. The Mindwerks repository is a fork. [It is referenced in one of its earliest commits](https://github.com/Mindwerks/XLEngine/commit/34edf014e6b4d0ffd1fd6b6817f90d0d8490c008)

cxong: yeah, but that doesn't mean the fork should be removed, since there are lots of commits after the fork and the fork is quite popular. I don't know more about the two projects but it seems that The Force Engine takes a different approach to XL Engine, so it's worth keeping both around, especially when the fork's developers haven't stated their intention with The Force Engine. It would be ok to expand on this in the comments; as it currently stands it is slightly inaccurate:
- "from the same developer" - not quite, the fork has other developers who have put in lots of commits
- "superseded" - the project that was superseded was the original project, not the fork

Phredreeke: The important part IMO is in regard to the supported games. At the moment it gives the impression of SW, Blood and Outlaws being supported.

Talking with people more knowledgable than me about Dark Forces, DarkXL has several game breaking bugs. Hence my insistence of directing people towards TFE (in which the whole game can be completed)

cxong (approved):

games/x.yaml
cxong: The page shows many indicators that TFE is a better project than DarkXL - status, development, and github stars, so it shouldn't be a big concern

PR #2365, opened 2023-12-07
Add Beasts and Bumpkins, SimCity 4, and FIFA 2001

Razzaline:


originals/s.yaml
```diff
@@ -505,6 +505,26 @@
...
+  - SimCity 4 Deluxe Edition
+  external:
+    website: 'https://www.ea.com/games/simcity/simcity-4'
+    wikipedia: SimCity 4
+  platforms:
+  - OSX
+  - Windows
+  meta:
+    genres:
+    - Real-Time Strategy
```
cxong, line 520: I think this genre should be removed. True it is a real time and also a strategy game, but the "RTS" term is used for a more specific kind of game, typically involving base building, unit control, and resource gathering.

Razzaline: That makes sense. I'll change that now. Genres can be a bit confusing regarding what 'technically' fits in one and what's generally understood to be part of the genre.

cxong (approved):

PR #2367, opened 2023-12-07
Add Dragon Alpha

Razzaline:


Razzaline:
Huh, is there a reason it requires an "external:" section? In this case it's an old abandonware game and doesn't have an official website anymore, so I wasn't sure what, if anything, to put in there.

originals/d.yaml
```diff
@@ -507,6 +507,16 @@
     themes:
     - Fantasy
 
+- name: Dragon Alpha
```
cxong, line 510: external is required - needs wikipedia or website

cxong (changes requested):

originals/d.yaml
Razzaline: Well, as far as a website is concerned, there's the GitHub repository for the game (https://github.com/jaredplumb/Dragon-Alpha), but that's the closest I have... I don't think its old page on the developer's website (Dracosoft) has been archived so I can't give a Wayback Machine link, and other than that there are just third-party websites that have the game uploaded or have information about it.

originals/d.yaml
```diff
@@ -507,6 +507,18 @@
...
+  external:
+    website: 'https://github.com/jaredplumb/Dragon-Alpha'
```
cxong, line 512:
```suggestion
    website: https://archive.org/details/tucows_233608_Dragon_Alpha
```

originals/d.yaml
cxong: I think a third party website is preferred. The repo does give good info though

games/d.yaml
```diff
@@ -1046,6 +1046,26 @@
...
+  - Dragon Alpha
+  type: official
+  repo: 'https://github.com/jaredplumb/Dragon-Alpha'
+  development: halted
+  status: playable
+  content: open
+  langs:
+  - C++
+  frameworks:
+  - DrawSprocket
```
cxong, line 1060:
```suggestion
  - Game Sprockets
```
Not familiar with this API but it sounds more general than DrawSprocket, similar to DirectX
Please also add this to dangerfile.js to silence the frameworks warning

cxong (changes requested):

games/d.yaml
Razzaline: Yeah, DrawSprocket is one component of Game Sprockets, which was an older Mac OS (between Classic and OSX) framework that meant to be an equivalent to DirectX. I wasn't sure if I should use "DrawSprocket" or  "Game Sprockets" because they only seemed to be using the DrawSprocket component, but I guess it would make sense in a similar situation to say "DirectX" instead of "DirectInput" or "DirectSound".

Also, should I add Game Sprockets to dangerfile.js as "Game Sprockets" or "GameSprockets"? The former seems to be used more by third parties (including Wikipedia), while the latter seems to be what Apple used themselves. Or am I able to add both versions?

cxong (approved):

PR #2374, opened 2023-12-13
Add Populous, Populous II, and open source projects related to the series

Razzaline:
Most were from https://github.com/OpenPop/notes but some were added from https://toksisitee.github.io/blog/Projects, https://github.com/PopRe, or incidental searching. Unfortunately, most projects haven't been touched for a long time, but maybe having them in the list can help anyone who might be interested in continuing them or creating their own.

I also had to add a few new frameworks to [dangerfile.js](https://github.com/Razzaline/osgameclones/blob/pr-populous/dangerfile.js) and add X68000 to platforms in [originals.yml](https://github.com/Razzaline/osgameclones/blob/pr-populous/schema/originals.yaml).


Razzaline:
I'll fix this mess >.<

PR #2445, opened 2024-02-02
Update Hurrican

midzer:
Hi,
just updating repo link and development.

Cheers
midzer

### DIFF
--- a/games/h.yaml
+++ b/games/h.yaml
@@ -526,7 +526,7 @@
   licenses:
   - As-is
   content: free
-  development: active
+  development: sporadic
   originals:
   - Turrican
   status: playable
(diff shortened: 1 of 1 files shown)


games/h.yaml
```diff
@@ -526,13 +526,13 @@
   licenses:
   - As-is
   content: free
-  development: halted
+  development: active
```
cxong, line 529:
```suggestion
  development: sporadic
```
last commit activity seems to be two months ago; based on OSGC definitions it should be sporadic

cxong (changes requested):

cxong (approved):

PR #2556, opened 2024-05-07
Add rsc-c

alarixnia:


cxong (changes requested):
YAML needs to be formatted correctly - each game is in a list element

PR #2563, opened 2024-05-12
Fix OpenHV screenshots

Mailaender:
as @discord does not want to be a free CDN anymore.

### DIFF
--- a/games/o.yaml
+++ b/games/o.yaml
@@ -1529,9 +1529,9 @@
 
 - name: OpenHV
   images:
-  - https://media.discordapp.net/attachments/388282819371204608/1043600071691862157/Screenshot_20221119_193803.png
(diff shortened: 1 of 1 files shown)


games/o.yaml
```diff
@@ -1554,7 +1554,8 @@
   repo: https://github.com/OpenHV/OpenHV
   type: similar
   url: https://openhv.net
-  updated: 2023-03-05
+  added: 2021-01-07
```
cxong, line 1557:
```suggestion
  added: 2020-12-30
```
in https://github.com/opengaming/osgameclones/commit/f6c62bceae91541fae988f1cd04ef97a9a04858e

cxong (changes requested):

PR #2606, opened 2024-06-17
Add some games

M0REKZ:
This adds 3 different things (2 of these also adds an original Game)

### DIFF
--- a/originals/t.yaml
+++ b/originals/t.yaml
@@ -54,23 +54,6 @@
   - OSX
   - Windows
 
-- name: Teeworlds
-  names:
-  - TeeWars
-  external:
-    website: 'https://www.teeworlds.com'
-    wikipedia: Teeworlds
-  platforms:
-  - Linux
-  - OSX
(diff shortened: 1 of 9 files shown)


originals/t.yaml
```diff
@@ -54,6 +54,23 @@
   - OSX
   - Windows
 
+- name: Teeworlds
```
cxong, line 57: OSGC doesn't list original open source games, so teeworlds should not be added

cxong: for original open source games consider contributing to projects like https://trilarion.github.io/opensourcegames

M0REKZ: What about DDraceNetwork? I was considering to add it here too, it's a teeworlds fork that is being distributed as a standalone Game on Steam

cxong: it shouldn't be added either imo

M0REKZ: Ok

PR #2799, opened 2024-12-01
Added Gikopoi reimplementations

CuloArdido:
Gikopoi was a multiplayer, online chatroom with avatars based on Shift_JIS art characters from 2channel ([here](https://namelessrumia.heliohost.org/w/doku.php?id=gikopoi) is most details about it). 

Some clarifications: 

* If the game had to be categorized within a genre, I think the most appropriate would be MMO (since it's basically like a Japanese version of Habbo) but that category is not available and considering it an MMORPG doesn't seem entirely correct to me. 
* Regarding the reimplementations, for the names I based on what appears on their respective websites, but if necessary you could also use the name they have in Github (Gikopoi2 and Gikopoi3).

### DIFF
--- a/games/g.yaml
+++ b/games/g.yaml
@@ -175,6 +175,37 @@
   video:
     youtube: lTtT9p9DVL8
 
+- name: Gikopoipoi
+  originals:
+  - Gikopoi
+  type: remake
+  repo: 'https://github.com/iccanobif/gikopoi2/'
+  url: 'https://gikopoipoi.net'
+  development: active
+  status: playable
+  langs:
+  - TypeScript
+  licenses:
+  - Unlicense
+  updated: '2024-12-01'
+  video:
+    youtube: TOMaQmWecKw
+
+- name: Gikopoi.com
+  originals:
+  - Gikopoi
+  type: remake
+  repo: 'https://github.com/153/gikopoi3'
+  url: 'https://gikopoi.com/'
+  development: active
+  status: playable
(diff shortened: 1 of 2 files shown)


games/g.yaml
```diff
@@ -175,6 +175,37 @@
...
+  type: remake
+  repo: 'https://github.com/iccanobif/gikopoi2/'
+  url: 'https://gikopoipoi.net'
+  development: active
+  status: playable
+  langs:
+  - TypeScript
+  licenses:
+  - Unlicense
+  updated: '2024-12-01'
```
cxong, line 190:
```suggestion
  added: '2024-12-06'
  updated: '2024-12-06'
```

games/g.yaml
```diff
@@ -175,6 +175,37 @@
...
+  repo: 'https://github.com/153/gikopoi3'
+  url: 'https://gikopoi.com/'
+  development: active
+  status: playable
+  langs:
+  - TypeScript
+  licenses:
+  - Custom
+  info: Fork of Gikopoipoi with some extra stuff.
+  updated: '2024-12-01'
```
cxong, line 207:
```suggestion
  added: '2024-12-06'
  updated: '2024-12-06'
```

153:
I would argue that gikopoi is more like a MUD than an MMO (it is not "massively" multiplayer) but otherwise good job, thanks for the link! -- gikopoi.com admin

PR #2923, opened 2025-03-12
Update twin-e

cxong:


xesf:
@cxong There's a more advanced port of this engine as part of the ScummVM https://github.com/scummvm/scummvm/tree/master/engines/twine

The original source code for both LBA1 and LBA2 have been publicly released https://github.com/2point21

PR #3021, opened 2025-05-18
Add O21, a remake of U95

ForNeVeR:
Long time no see, folks.

I am glad to present my (and, of course, of other project benefactors) latest invention — a "semi-playable" remake of a Windows game from 1995!

The license list is a bit long-ish, but that's because we are accurately tracking licenses from each used free asset – fonts, sound fonts, even code of conduct (that's not including the original's game abandonware/commercial content, of course).


ForNeVeR:
Not sure why does the CI fail. Is it a problem in my PR?

PR #3087, opened 2025-06-13
Add  Xash3D FWGS

ju5:


cxong:
From what I understand, we want to avoid adding xash3d related projects, as they are not legal code. See

https://github.com/opengaming/osgameclones/pull/92
https://github.com/opengaming/osgameclones/commit/70e0f667580ed11d5fd0189edc90344d794b5a78
https://github.com/FWGS/xash3d-fwgs/blob/c90d1100965ec5a072d95c0c257f853a6673ca74/common/hltv.h#L3

PR #3108, opened 2025-06-21
Update SR

ju5:
Refer project's [README](https://github.com/M-HT/SR)

### DIFF
--- a/originals/b.yaml
+++ b/originals/b.yaml
@@ -146,7 +146,7 @@
     genres:
     - Arcade
 
-- name: 'Battle Isle 2220: Shadow of the Emperor'
+- name: Battle Isle 2220
   external:
     website: >-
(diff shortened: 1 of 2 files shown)


originals/b.yaml
```diff
@@ -146,6 +146,17 @@
     genres:
     - Arcade
 
+- name: 'Battle Isle 2220: Shadow of the Emperor'
```
```suggestion
- name: Battle Isle 2220
```
let's use the shorter name

ju5: I noticed other original games like another original game by SR "Septerra Core: Legacy of the Creator" use full names with subtitles. Wouldn't shortening just this entry create inconsistency in our naming pattern?
Also, if we change to a short name here, should we also update the originals field in the SR entry to match?

cxong: yes, I think septerra core should use the shorter name

cxong (approved):

PR #3134, opened 2025-06-29
Add Kizare

ju5:


games/k.yaml
```diff
@@ -226,6 +226,26 @@
...
+  url: >-
+    https://web.archive.org/web/20110427084745/http://www.sdltutorials.com/sdl-games/#:~:text=kizare
+  development: halted
+  status: semi-playable
+  langs:
+  - C++
+  frameworks:
+  - SDL
+  - OpenGL
+  - SpiderMonkey
```
cxong:
```suggestion
```
I don't think spidermonkey is relevant here, it is basically a JS engine meaning it runs plain javascript, and is relevant only for packaging purposes and not for development

games/k.yaml
```diff
@@ -226,6 +226,26 @@
...
 
+- name: Kizare
+  originals:
+  - 'Legend of Zelda - A Link to the Past'
+  type: similar
+  repo: 'https://code.google.com/archive/p/kizare/'
+  url: >-
+    https://web.archive.org/web/20110427084745/http://www.sdltutorials.com/sdl-games/#:~:text=kizare
+  development: halted
+  status: semi-playable
```
cxong: I can't see evidence for this game being in any playable state, I had a look at the source and couldn't find any game data nor game logic code, so it seems this project never got off the ground. I'm not really sure about adding this game to OSGC, doesn't seem that interesting to me

ju5: No problem, if you think it's not appropriate to add, feel free to close the PR. I've changed the status to unplayable.

PR #3143, opened 2025-07-02
Add Mr. Drillux

ju5:


games/m.yaml
```diff
@@ -1156,6 +1156,32 @@
   - https://www.artsoft.org/mirrormagic/screenshots/infoscreen.png
   - https://www.artsoft.org/mirrormagic/screenshots/editor.png
 
+- name: Miss Driller
+  originals:
+  - Mr. Driller
+  type: clone
+  repo: >-
+    https://web.archive.org/web/20060304175723/http://imrtechnology.ngemu.com/downloads/driller2src.rar
```
```suggestion
    https://github.com/mthuurne/mrdrillux
```
this is a port but it is more up to date

games/m.yaml
```diff
@@ -1156,6 +1156,32 @@
...
+    https://web.archive.org/web/20060308054155/http://www.dcemu.co.uk/missdriller.shtml
+  development: halted
+  status: playable
+  content: free
+  langs:
+  - C
+  frameworks:
+  - SDL
+  licenses:
+  - As-is
```
```suggestion
  - LGPL2
```
based on JP readme

games/m.yaml
```diff
@@ -1156,6 +1156,32 @@
   - https://www.artsoft.org/mirrormagic/screenshots/infoscreen.png
   - https://www.artsoft.org/mirrormagic/screenshots/editor.png
 
+- name: Miss Driller
```
```suggestion
- name: Mr. Drillux
```
confusingly the game was ported a few times with name changes each time

ju5: Changed. Should the url field and even the YouTube ID be deleted?

cxong: I think it's ok to keep them; these ports tend to be pretty much the same game so these are still representative

PR #3164, opened 2025-07-06
Add Insanerz Shooter

ju5:
Reference https://os4depot.net/?function=showfile&file=game/shmup/insanerz_shooter.lha

### DIFF
--- a/games/i.yaml
+++ b/games/i.yaml
@@ -128,6 +128,27 @@
   updated: 2020-08-31
   url: https://inexor.org/
 
+- name: Insanerz Shooter
+  originals:
+  - Galaga
+  type: clone
+  repo: 'https://code.google.com/archive/p/insanerzshooter/'
+  url: >-
+    https://web.archive.org/web/20090303121207/http://insanerzshooter.googlepages.com/
+  development: halted
+  status: playable
+  content: open
+  langs:
+  - C++
+  frameworks:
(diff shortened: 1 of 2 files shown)


games/i.yaml
```diff
@@ -128,6 +128,27 @@
   updated: 2020-08-31
   url: https://inexor.org/
 
+- name: Insanerz Shooter
+  originals:
+  - Galaga
+  type: clone
+  repo: 'https://code.google.com/archive/p/insanerzshooter/'
```
cxong:
```suggestion
  repo: https://github.com/marciofrayze/insanerzshooter
```

cxong (changes requested):
I don't think this game should be added; it appears to be a very simple shooting game where randomly spawned enemies fly down the screen https://www.youtube.com/watch?v=scZQE4cdKH0 and therefore it is not a clone of galaga. If it is a clone of a game it would be too simple so I don't think it's notable enough

ju5:
> I don't think this game should be added; it appears to be a very simple shooting game where randomly spawned enemies fly down the screen https://www.youtube.com/watch?v=scZQE4cdKH0 and therefore it is not a clone of galaga. If it is a clone of a game it would be too simple so I don't think it's notable enough

Thanks for the thorough review, You can close this if it’s not suitable.

PR #3178, opened 2025-07-13
Add No Gravity

ju5:


cxong (commented):
I don't think this game should be added, as I don't see how it is a clone of specifically wing commander. It looks similar to many space combat games

ju5:
> I don't think this game should be added, as I don't see how it is a clone of specifically wing commander. It looks similar to many space combat games

https://www.wcnews.com/news/update/7728 mentions:

> No Gravity is a space shooter in 3D ' a la Wing Commander'.

Or, there is another option (See https://www.osnews.com/story/9736/nogravity-game-open-sourced/):
```yaml
- name: No Gravity
  originals:
  - SpaceGirl
  type: official
```
What are your thoughts?

cxong:
I see there's another page which mentions this is similar to wing commander https://www.themacattic.com/title/840a93f81f734981-no-gravity-aka-space-girl
I guess it's ok to add since it appears it is not just one random writer saying so

originals/w.yaml
```diff
@@ -207,6 +207,17 @@
     themes:
     - Sci-Fi
 
+- name: Wing Commander series
+  external:
+    wikipedia: Wing Commander (franchise)
+  meta:
+    genres:
+    - Action
+    - Adventure
```
cxong:
```suggestion
```

cxong (approved):

PR #3181, opened 2025-07-14
Add The Legend of Zelda: A Link to the Dream

ju5:
Reference https://www.solarus-games.org/games/the-legend-of-zelda-a-link-to-the-dream/

### DIFF
--- a/originals/l.yaml
+++ b/originals/l.yaml
@@ -195,9 +195,7 @@
     - Action
     - Adventure
 
-- name: "The Legend of Zelda: Link’s Awakening"
-  names:
-  - "The Legend of Zelda: Link's Awakening DX"
+- name: "The Legend of Zelda: Link's Awakening"
   external:
     website: >-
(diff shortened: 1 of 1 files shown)


originals/l.yaml
```diff
@@ -195,6 +195,22 @@
     - Action
     - Adventure
 
+- name: "The Legend of Zelda: Link’s Awakening"
```
cxong:
```suggestion
- name: "The Legend of Zelda: Link's Awakening"
```

originals/l.yaml
```diff
@@ -195,6 +195,22 @@
...
+  names:
+  - "The Legend of Zelda: Link's Awakening DX"
```
cxong:
```suggestion
```
the names are very similar so no need to include it

originals/l.yaml
ju5: You're really observant. That was my mistake.

cxong (approved):

PR #3192, opened 2025-07-19
Add xclannad

ju5:


ju5:
@cxong Added a missing original game, Please take a look.

PR #3198, opened 2025-07-21
Add Worm Wars

ju5:


games/w.yaml
```diff
@@ -703,6 +703,26 @@
   video:
     youtube: 9VaW2NVxpVg
 
+- name: Worm Wars
+  originals:
+  - Tron
+  - Snake
+  type: similar
```
```suggestion
  type: clone
```

games/w.yaml
```diff
@@ -703,6 +703,26 @@
   video:
     youtube: 9VaW2NVxpVg
 
+- name: Worm Wars
+  originals:
+  - Tron
+  - Snake
```
```suggestion
```
tron is already a snake like game so this is redundant

PR #3206, opened 2025-07-24
Add Children of Solarus

ju5:


games/c.yaml
```diff
@@ -616,6 +616,33 @@
   type: remake
   updated: 2019-10-26
 
+- name: Children of Solarus
+  originals:
+  - Legend of Zelda - A Link to the Past
+  type: similar
+  repo: 'https://gitlab.com/solarus-games/children-of-solarus'
```
```suggestion
  repo: https://gitlab.com/solarus-games/games/children-of-solarus
```

games/c.yaml
```diff
@@ -616,6 +616,33 @@
   type: remake
   updated: 2019-10-26
 
+- name: Children of Solarus
+  originals:
+  - Legend of Zelda - A Link to the Past
+  type: similar
```
```suggestion
  type: clone
```

ju5: Children of Solarus is a remake of Mystery of Solarus DX (a fan-made sequel to A Link to the Past), so I think it should be classified as 'similar' to A Link to the Past.

cxong: I think Mystery of Solarus DX should be a remake, as it uses the same assets, and the game play is the same, even though the story and levels are different
For this game, since the only similarity with the original game is the gameplay, it should be a clone instead

PR #3230, opened 2025-07-29
Add Sword of Fargoal remake

ju5:


games/s.yaml
```diff
@@ -3188,6 +3188,24 @@
   updated: 2023-11-11
   url: http://www.schillmania.com/survivor/
 
+- name: Sword of Fargoal remake
+  originals:
+  - Sword of Fargoal
+  type: remake
+  repo: 'https://archive.org/download/fargoal/fargoal.zip/'
```
```suggestion
```
don't want to link directly to a file

ju5: Adding `/` at the end of an `archive.org` zip file link makes it *browse the file contents* instead of triggering a direct download.

cxong: that's cool
it is not needed however because the page already provides a download link, and this page only allows viewing file listings and downloading individual files, and does not allow viewing the files without downloading, so it is not very useful

cxong (approved):

PR #3235, opened 2025-07-30
Add Xpuyopuyo

ju5:


games/x.yaml
```diff
@@ -535,6 +535,26 @@
   updated: 2022-02-04
   url: https://xpilot.sourceforge.net
 
+- name: Xpuyopuyo
+  originals:
+  - Puyo Puyo Series
+  type: clone
+  url: 'https://chaos2.org/xpuyopuyo/'
```
cxong: this site seems to be down
https://downforeveryoneorjustme.com/chaos2.org?proto=https

ju5: I can access this website, but it is currently unavailable in some regions.
https://ping.pe/chaos2.org

games/x.yaml
```diff
@@ -535,6 +535,26 @@
...
+  development: halted
+  status: semi-playable
+  content: open
+  langs:
+  - C
+  frameworks:
+  - GTK
+  licenses:
+  - GPL1
```
```suggestion
  - GPL2
```

games/x.yaml
cxong: it seems to be unavailable in many regions, so it's not a good url. We should use alternate ones for it and images, like https://archive.org/details/tucows_50468_Xpuyopuyo for example

PR #3236, opened 2025-07-30
Add XBoing

ju5:
Reference https://packages.altlinux.org/en/p11/srpms/xboing/specfiles/

### DIFF
--- a/games/x.yaml
+++ b/games/x.yaml
@@ -129,7 +129,7 @@
 - name: XBoing
   originals:
   - Breakout
-  type: similar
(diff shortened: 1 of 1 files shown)


games/x.yaml
```diff
@@ -126,6 +126,31 @@
   updated: 2014-08-15
   url: https://xbak.sourceforge.net
 
+- name: XBoing
+  originals:
+  - Breakout
+  type: similar
```
```suggestion
  type: clone
```

cxong (approved):

PR #3241, opened 2025-07-31
Add Flappy Bird for Gymnasium

ju5:


games/f.yaml
```diff
@@ -329,7 +329,28 @@
...
+
+- name: Flappy Bird for Gymnasium
+  originals:
+  - Flappy Bird
+  type: clone
+  repo: 'https://github.com/markub3327/flappy-bird-gymnasium'
+  feed: 'https://github.com/markub3327/flappy-bird-gymnasium/releases.atom'
+  development: halted
+  status: playable
+  content: open
```
cxong: what is the provenance of the game's assets? afaict they look identical to the original game's assets, which I'm not aware of being open source
https://github.com/markub3327/flappy-bird-gymnasium/tree/main/flappy_bird_gymnasium/assets

ju5: The assets were likely extracted from the original game, what value would be more appropriate for the content field?

cxong: I think it's ok to say it's free then, since a lot of clones also use the same assets

PR #3242, opened 2025-08-01
Add BuildGDX

ju5:
Reference https://m210.duke4.net/index.php/42

### DIFF
--- a/games/b.yaml
+++ b/games/b.yaml
@@ -613,7 +613,7 @@
   frameworks:
   - libGDX
   licenses:
-  - Custom
+  - GPL3
   added: 2025-08-02
   updated: 2025-08-02
   video:
--- a/games/b.yaml
+++ b/games/b.yaml
@@ -1061,6 +1061,33 @@
   updated: 2022-03-13
   url: http://bub-n-bros.sourceforge.net/
 
+- name: BuildGDX
+  originals:
+  - Blood
+  - Duke Nukem 3D
+  - Redneck Rampage
+  - Shadow Warrior
+  - Witchaven
+  - 'Witchaven II: Blood Vengeance'
+  - "Rock'n Shaolin: Legend of Seven Paladins 3D"
+  - TekWar
+  - Powerslave
+  type: remake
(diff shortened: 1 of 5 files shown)


cxong (changes requested):
I believe the way this author (m210) structured these ports is that each game has its own project (for example dukegdx is already on OSGC) but they all use buildgdx as a dependency. I think it's better if we add the individual ports rather than buildgdx, this allows us to track things like screenshots/videos and playability status for each port

games/b.yaml
```diff
@@ -598,6 +598,27 @@
...
+  feed: https://gitlab.com/m210/BloodGDX.atom
+  development: sporadic
+  status: playable
+  content: commercial
+  langs:
+  - Java
+  frameworks:
+  - libGDX
+  licenses:
+  - Custom
```
```suggestion
  - GPL3
```

games/l.yaml
```diff
@@ -1171,6 +1171,27 @@
...
+  feed: 'https://gitlab.com/m210/LSPGDX.atom'
+  development: sporadic
+  status: playable
+  content: commercial
+  langs:
+  - Java
+  frameworks:
+  - libGDX
+  licenses:
+  - Custom
```
cxong:
```suggestion
  - GPL3
```

games/w.yaml
```diff
@@ -60,6 +60,27 @@
...
+  feed: https://gitlab.com/m210/wanggdx.atom
+  development: sporadic
+  status: playable
+  content: commercial
+  langs:
+  - Java
+  frameworks:
+  - libGDX
+  licenses:
+  - Custom
```
```suggestion
  - GPL3
```

originals/r.yaml
```diff
@@ -414,6 +414,26 @@
...
+    website: 'https://www.mobygames.com/game/56679/euijeok-im-kkeokjeong/'
+  platforms:
+  - DOS
+  meta:
+    genres:
+    - Action
+    - FPS
+    - RPG
+    subgenres:
+    - Action RPG
```
cxong:
```suggestion
    genres:
    - FPS
```

cxong (approved):

PR #3245, opened 2025-08-02
Add rustcraft

ju5:


games/r.yaml
```diff
@@ -1299,6 +1299,36 @@
...
+  - Bevy
+  licenses:
+  - PD
+  - CC-BY-SA
+  - MIT
+  added: 2025-08-02
+  updated: 2025-08-02
+  images:
+  - 'https://github.com/c2i-junia/rustcraft/raw/main/docs/preview1.png'
+  - 'https://github.com/c2i-junia/rustcraft/blob/main/docs/preview2.png'
```
```suggestion
  - https://github.com/c2i-junia/rustcraft/blob/main/docs/preview2.png?raw=true
```

games/r.yaml
```diff
@@ -1299,6 +1299,36 @@
...
+  - >-
+    https://github.com/c2i-junia/rustcraft/blob/main/docs/images/preview1.png?raw=true
+  - >-
+    https://github.com/c2i-junia/rustcraft/blob/main/docs/images/preview2.png?raw=true
```
```suggestion
```

ju5: I standardized the image URL format. These are four different screenshots, why should the last two be deleted?

cxong: the deleted ones are the same as the first two

ju5: I believe there might be a small misunderstanding — the last two image paths contain an additional images/ subdirectory, so they are actually different from the first two.

PR #3247, opened 2025-08-03
Add xspacewarp

ju5:


originals/s.yaml
cxong, line 1: this entry seems wrong, it should be time trek https://www.mobygames.com/game/49960/time-trek/ and also not Genesis (Mega Drive)

ju5: According to the [readme](https://www.ibiblio.org/pub/micro/pc-stuff/Linux/games/arcade/xspacewarp-1.2.README), Space Warp is the commercial version of Time Trek. Which one is more suitable as the original game?

cxong: ok in that case let's keep it as Space Warp, however I don't think Time Trek should be considered an alternate name

cxong (approved):

PR #3250, opened 2025-08-04
Add Left4DeadRL

ju5:


games/l.yaml
cxong, line 1: I don't think this game should be added, its relation to l4d is extremely tenuous. It's not the same genre, I don't see any similarity theme-wise apart from being a zombie themed game.

cxong (changes requested):

games/l.yaml
ju5: I have no objection. If it shouldn't be added, feel free to close it.

mcandre: left4deadrl is a minimalist *clone* of Left4Dead based on the same premise of AI director driven entertainment.

The game takes direct inspiration from the other.

However, there is no particular joy in the game. I never got around to adding story, items, or balancing difficulty. It's more of a bare prototype and an experimentation with console graphics.

PR #3254, opened 2025-08-05
Add Bomb-her-man

ju5:


ju5:
I noticed that this game is already hosted in two software repositories, which I believe demonstrates its notability to some extent.:
https://www.freshports.org/games/bombherman
https://github.com/fd00/yacp/tree/master/Bomb-her-man

PR #3261, opened 2025-08-07
Add Dark Well

ju5:


games/d.yaml
```diff
@@ -186,7 +186,27 @@
   url: http://dark-oberon.sourceforge.net/
   video:
     youtube: pgFWAeM4Rm8
-  
+
+- name: Dark Well
+  originals:
+  - Columns
+  type: similar
```
```suggestion
  type: clone
```
I can't tell what this game is like and how similar it is so I'm assuming it's very similar - note a lot of people don't like referring to their own games as clones

PR #3266, opened 2025-08-08
Add Brick Bird

ju5:


games/b.yaml
```diff
@@ -940,6 +940,30 @@
...
+  repo: 'https://github.com/Siiir/brick_bird/'
+  url: 'https://siiir.github.io/brick_bird/'
+  feed: 'https://github.com/Siiir/brick_bird/releases.atom'
+  development: halted
+  status: playable
+  content: free
+  langs:
+  - Rust
+  licenses:
+  - Custom
```
```suggestion
  - As-is
```
the author calls it an "all rights reserved license" but this is an oxymoron, we can consider it all rights reserved i.e. no license

cxong (approved):

PR #3270, opened 2025-08-08
Add Boing!

ju5:


games/b.yaml
```diff
@@ -1085,6 +1085,24 @@
...
+  url: 'https://www.ibiblio.org/pub/micro/pc-stuff/Linux/games/arcade/boing-0.1.lsm'
+  development: halted
+  status: playable
+  content: open
+  langs:
+  - C
+  frameworks:
+  - SVGALib
+  licenses:
+  - GPL1
```
```suggestion
  - GPL2
```
I can't find any mention of licenses in the source files, so I'm inclined to believe it's actually GPL2 as it is much more common than GPL1

PR #3330, opened 2025-08-25
Add 2048.c

ju5:


mevdschee:
I'm more proud of my minesweeper clone: https://github.com/mevdschee/raylib-go-mines

But I like you addition. Thank you.

PR #3351, opened 2025-09-01
Add XP Minesweeper Classic

Kurtsley:


games/x.yaml
```diff
@@ -683,6 +683,31 @@
   updated: 2022-02-04
   url: https://xpilot.sourceforge.net
 
+- name: XP Minesweeper Classic
+  originals:
+  - Minesweeper
+  type: clone
```
```suggestion
  type: remake
```

games/x.yaml
```diff
@@ -683,6 +683,31 @@
...
+  repo: 'https://github.com/Kurtsley/XP-Minesweeper-Classic'
+  development: active
+  status: playable
+  content: open
```
```suggestion
  content: free
```
there is art owned by microsoft, I don't think they can be open source unless they have released them explicitly so

Kurtsley: Indeed.

cxong (approved):

PR #3381, opened 2025-09-09
Add original "Starbound" and game "OpenStarbound"

Jeddunk:


originals/s.yaml
```diff
@@ -1508,6 +1508,25 @@
...
     - Sci-Fi
 
+- name: Starbound
+  external:
+    website: 'https://playstarbound.com/'
+    wikipedia: Starbound
+  platforms:
+  - Linux
+  - OSX
+  - Windows
```
cxong, line 1518:
```suggestion
  - Windows
  - Xbox One
  - Xbox Series S/X
```

games/o.yaml
```diff
@@ -2690,6 +2690,29 @@
   type: remake
   updated: 2021-11-19
 
+- name: OpenStarbound
```
cxong, line 2693: it seems this project contains leaked source code and is thus not legal, we should not add it to OSGC

cxong (changes requested):

games/o.yaml
Jeddunk: Yes, as far as I'm aware, OpenStarbound is based on the leaked source code of the last version of the game.

cxong:
don't want to add OpenStarbound

PR #3382, opened 2025-09-09
Add Wordle CLI

ju5:


games/w.yaml
```diff
@@ -725,6 +725,26 @@
   video:
     youtube: 9VaW2NVxpVg
 
+- name: Wordle CLI
+  originals:
+  - Wordle
+  type: clone
+  repo: 'https://github.com/nimblebun/wordle-cli'
+  url: 'https://aur.archlinux.org/packages/wordle-cli-bin'
```
```suggestion
```
this is only relevant for arch users and the game is not arch exclusive so isn't really good for a URL

cxong (approved):

PR #3386, opened 2025-09-11
Add Glider for Apple II

colinleroy:


games/g.yaml
```diff
@@ -259,6 +259,27 @@
   added: 2025-08-26
   updated: 2025-08-26
 
+- name: Glider for Apple II
+  originals:
+  - Glider
+  type: remake
+  repo: 'https://github.com/colinleroy/a2tools/tree/master/src/glider'
```
cxong:
```suggestion
  repo: https://github.com/colinleroy/a2tools
```
osgc doesn't handle monorepos well; instead we can have a single project associated with multiple original games

originals/g.yaml
```diff
@@ -160,6 +160,15 @@
     - Medieval
     - Fantasy
 
+- name: Glider
+  external:
+    wikipedia: 'https://en.wikipedia.org/wiki/Glider_(video_game)'
```
cxong:
```suggestion
    wikipedia: Glider (video game)
```

originals/g.yaml
```diff
@@ -160,6 +160,15 @@
...
+  platforms:
+  - Classic Mac OS
```
cxong, line 167:
```suggestion
  - Classic Mac OS
  - Windows
```

games/g.yaml
```diff
@@ -259,6 +259,27 @@
   added: 2025-08-26
   updated: 2025-08-26
 
+- name: Glider for Apple II
```
cxong, line 262:
```suggestion
- name: a2tools
```
should be named after the monorepo; multiple entries can be added under originals

cxong (changes requested):

games/g.yaml
cxong: my bad I think in this case because the URL is different it can be added like this

PR #3404, opened 2025-09-15
Add eshaansmart

ju5:


games/e.yaml
```diff
@@ -616,6 +616,23 @@
...
   - https://www.etlegacy.com/storage/shots/shot0005.jpeg
 
+- name: eshaansmart
+  originals:
+  - Monkey Mart
+  type: clone
+  repo: 'https://github.com/CodeMiner11-py/eshaansmart'
+  url: 'https://pypi.org/project/eshaansmart/'
+  development: sporadic
+  status: semi-playable
```
cxong: how is it semi playable? I can't see any indication of if the game runs or works

ju5: Based on the information that this game was developed by a 12-year-old boy, I suspect its maturity level is not high. Combined with your input, I have now changed the status to unplayable.

Issue #2969, opened 2025-04-21
Add Friday Night Funkin' Rewritten


ju5:
The [original game](https://ninja-muffin24.itch.io/funkin) is also [open source](https://github.com/FunkinCrew/funkin), and I'm not sure it meets the inclusion criteria.
```
name: "Friday Night Funkin' Rewritten"
originals:
- "FRIDAY NIGHT FUNKIN'"
type: remake
repo: 'https://github.com/HTV04/funkin-rewritten'
development: sporadic
status: playable
content: open
langs:
- Lua
frameworks:
- LÖVE
licenses:
- GPL3
info: "Optimized rewrite of Friday Night Funkin' built on LÖVE."
added: '2025-04-22'
updated: '2025-04-22'
images:
- 'https://raw.githubusercontent.com/htv04/funkin-rewritten/main/images/logo.png'
```

cxong:
I think in this case we don't want to add to OSGC. There are games where the original game was closed source first but then opened later. There are also games where the original is open source, back when we were more lenient, but would not have been added today, and for legacy reasons we're keeping them.

PR #3415, opened 2025-09-18
Add Candy Crisis

ju5:


games/c.yaml
```diff
@@ -120,6 +120,31 @@
   updated: 2020-02-08
   url: https://bitbucket.org/dalerank/caesaria/wiki/Home
 
+- name: Candy Crisis
+  originals:
+  - Puyo Puyo Series
```
```suggestion
  - Candy Crisis
```
this should be considered a source port (remake)

games/c.yaml
```diff
@@ -120,6 +120,31 @@
...
+  type: clone
+  repo: 'https://github.com/jorio/CandyCrisis'
+  url: 'https://candycrisis.sourceforge.net/'
```
```suggestion
```

cxong (changes requested):

games/c.yaml
ju5: I’d like to clarify the standards of OSGC based on your previous comment. In [#2969](https://github.com/opengaming/osgameclones/issues/2969#issuecomment-2819683580), you mentioned that OSGC does not accept clones of games that the original is open source. This seems inconsistent with treating "Candy Crisis" as the original game.
According to past practices in OSGC, if game B is an open-source clone of game A, and game C is forked from game B, then both B and C should list A as the original. For example, in the current OSGC database, both "XGalaga" and "XGalaga++" list "Galaga" as their original.
I considered the GitHub repo a port of the Sourceforge project and merged them for brevity (similar to [#3143](https://github.com/opengaming/osgameclones/pull/3143#discussion_r2185018598)). If this is incorrect, I can separate them into two entries.

cxong: Candy Crisis was originally not open source, see https://gamicus.fandom.com/wiki/Candy_Crisis

PR #3418, opened 2025-09-19
Add minecurses

ju5:


games/m.yaml
```diff
@@ -1053,6 +1053,22 @@
   video:
     youtube: 4O0_-1NaWnY
 
+- name: minecurses
+  originals:
+  - Minesweeper
+  type: clone
+  repo: 'https://github.com/EliteTK/minecurses'
+  development: halted
+  status: semi-playable
```
cxong, line 1062:
```suggestion
  status: unplayable
```
according to the readme it looks very early

EliteTK: You can play it, I just never implemented a win state. But in default terminal sizes, it generates enough squares and mines that, most people get bored before they get that far. :P

cxong (approved):

PR #3437, opened 2025-09-22
Add Eastern Front (1941)

ju5:


games/e.yaml
```diff
@@ -33,6 +33,30 @@
   type: remake
   updated: 2021-01-24
 
+- name: Eastern Front (1941)
+  originals:
+  - Eastern Front (1941)
+  type: official
+  repo: 'https://github.com/scttgs0/EasternFront1941'
```
cxong:
```suggestion
```
this looks like a port

ju5: I switched to another one.

PR #3440, opened 2025-09-22
Add C2048

ju5:


cxong (changes requested):
I don't think this should be added - it is a library not a game

ju5:
I see. Feel free to close it then.

PR #3442, opened 2025-09-22
Add Bansoko

ju5:


originals/s.yaml
```diff
@@ -855,6 +855,18 @@
   external:
     wikipedia: Snipes (video game)
 
+- name: Soko-Ban
```
cxong, line 858: Sokoban is already available

ju5: According to readme and comparison level screenshots, Bansoko is a direct clone of Soko-Ban (https://www.mobygames.com/game/1715/soko-ban/), and I think Soko-Ban is more suitable as the original game.

PR #3489, opened 2025-10-02
Add Xonix

ju5:


originals/x.yaml
```diff
@@ -71,6 +71,17 @@
...
+- name: Xonix
+  external:
+    website: 'https://www.mobygames.com/game/46702/xonix/'
+  platforms:
+  - DOS
+  meta:
+    genres:
+    - Action
+    themes:
+    - Fantasy
```
cxong:
```suggestion
    - Abstract
```
I think the mobygames entry is misclassified

PR #3519, opened 2025-10-06
Add Scorch Went Bonkers

ju5:


games/s.yaml
```diff
@@ -156,6 +156,28 @@
   images:
   - https://raw.githubusercontent.com/fr0zi/sccopter/master/Refs/milestone_1_full_terrain_data.png
 
+- name: Scorch Went Bonkers
+  originals:
+  - Scorched Earth
+  type: clone
```
```suggestion
  type: similar
```
the game is real time so the gameplay is very different

PR #3545, opened 2025-10-10
Add RA2Web React

ju5:


games/r.yaml
```diff
@@ -1,3 +1,31 @@
...
+  langs:
+  - TypeScript
+  frameworks:
+  - React
+  - three.js
+  - WebGL
+  licenses:
+  - GPL3
+  info: >-
+    Developed based on the analysis of the Chinese version RA2WEB of the Chrono
```
```suggestion
    Developed based on decompiling the Chinese version RA2WEB of the Chrono
```

games/r.yaml
```diff
@@ -1,3 +1,31 @@
...
+  status: semi-playable
+  content: commercial
+  langs:
+  - TypeScript
+  frameworks:
+  - React
+  - three.js
+  - WebGL
+  licenses:
+  - GPL3
```
```suggestion
  - As-is
```
since it is based on closed source code it can't be made open source without original author's permission

PR #3547, opened 2025-10-11
Add 2048-puzzle

ju5:


rhoit:
[ASCII-board](https://github.com/bekar/ASCII-board) is the simple library to calculate board size and create ASCII art font, it was originally developed for 2048 and later expanded to few other TTY games.

PR #3600, opened 2025-10-19
Add Monkey Bubble

ju5:


games/m.yaml
```diff
@@ -1788,6 +1788,35 @@
   images:
   - https://identicalsoftware.com/mojotron/screenshot1.png
 
+- name: Monkey Bubble
+  originals:
+  - Puzzle Bobble
+  type: clone
+  repo: >-
+    https://web.archive.org/web/20081008073432/http://www.monkey-bubble.org/downloads
```
```suggestion
  repo: https://github.com/herzi/monkey-bubble
```

games/m.yaml
```diff
@@ -1792,8 +1792,7 @@
   originals:
   - Puzzle Bobble
   type: clone
-  repo: >-
-    https://web.archive.org/web/20081008073432/http://www.monkey-bubble.org/downloads
+  repo: https://github.com/herzi/monkey-bubble
```
herzi: I think this is a better source: https://gitlab.gnome.org/Archive/monkey-bubble

MB never was “my” game, I just contributed a few times.

herzi (changes requested):

PR #3666, opened 2025-10-29
Add Ladder repo

ju5:


games/l.yaml
```diff
@@ -78,8 +78,9 @@
   status: playable
   type: remake
   added: 2015-05-09
-  updated: 2015-05-10
+  updated: 2025-10-30
   url: https://ostermiller.org/ladder/
+  repo: https://sourceforge.net/projects/laddergame/
```
cxong, line 83: this doesn't have the source code

ju5: The homepage does not publish the source code either (https://ostermiller.org/ladder/download.html). As stated in the info section, the source code can be extracted from the JAR file.

PR #3688, opened 2025-11-01
Add Lode Runner SVGA

ju5:


cxong (changes requested):
doesn't look like the release contains source code

PR #3698, opened 2025-11-02
Add Airline Tycoon

ju5:


games/a.yaml
```diff
@@ -353,6 +353,24 @@
   images:
   - 'https://oldschoolprg.x10.mx/images/prj_airball0.png'
 
+- name: Airline Tycoon
+  originals:
+  - Airline Tycoon
+  type: official
```
```suggestion
  type: remake
```
it contains significant changes from original so it should be a remake

PR #3702, opened 2025-11-03
Add River Raid (VHDL)

ju5:


games/r.yaml
```diff
@@ -993,6 +993,22 @@
   - https://i.stack.imgur.com/Nz443.png
   - https://icculus.org/~ravage/rott/rott_installer.jpg
 
+- name: River Raid (2020)
```
```suggestion
- name: River Raid (VHDL)
```
I think this is more notable

PR #3710, opened 2025-11-04
Add MicroWar

ju5:


games/m.yaml
```diff
@@ -1032,6 +1032,31 @@
...
+  url: 'https://microwar.sourceforge.net/'
+  feed: 'https://sourceforge.net/p/microwar/news/feed.rss'
+  development: halted
+  status: playable
+  content: open
+  langs:
+  - Python
+  frameworks:
+  - pygame
+  - SDL
```
```suggestion
```
it should be using SDL indirectly via pygame

PR #3711, opened 2025-11-04
Add AB3D-RTG

ju5:


originals/a.yaml
```diff
@@ -268,6 +268,21 @@
...
+  external:
+    website: 'https://www.mobygames.com/game/12217/alien-breed-3d/'
+    wikipedia: Alien Breed 3D
+  platforms:
+  - Amiga
+  meta:
+    genres:
+    - FPS
+    subgenres:
+    - Run and gun
```
cxong:
```suggestion
```
run and gun is a subgenre of shmups, not FPS

PR #3714, opened 2025-11-05
Add Do The Same

ju5:


originals/d.yaml
```diff
@@ -499,6 +499,15 @@
     themes:
     - Fantasy
 
+- name: Do The Same
+  external:
+    website: 'https://web.archive.org/web/20160307152033/http://dothesame.jagware.org/'
+  platforms:
+  - Atari
```
cxong: should be jaguar

ju5: [Jaguar](https://en.wikipedia.org/wiki/Atari_Jaguar) is not in the platforms list.

cxong: it can be added to the schema as Jaguar

cxong (approved):

PR #3742, opened 2025-11-07
Add Checkered Flag

ju5:


games/c.yaml
```diff
@@ -678,6 +678,22 @@
   type: remake
   updated: 2019-10-26
 
+- name: Checkered Flag
+  originals:
+  - Checkered Flag
+  type: official
+  repo: >-
+    https://www.jagware.org/jag_uploads/dev/curt_vendel_jagfiles/#:~:text=Checkered%20Flag
+  url: 'https://www.youtube.com/watch?v=WYY_Q1tZOWM&t=327s'
```
```suggestion
```
I don't think we should link to a video for the URL

PR #3745, opened 2025-11-08
Add Minegauler

ju5:


LewisGaul:
> - 🏗 Minegauler uses "Qt" as a framework, but doesn't have C++ in its languages.

Maybe "PyQt" should be specified instead?

cxong (approved):

PR #3746, opened 2025-11-08
Add Alpha Waves Classic

ju5:


originals/c.yaml
```diff
@@ -680,6 +680,18 @@
   - DOS
   - Windows
 
+- name: Continuum
```
cxong: this is the same game as alpha waves

cxong (changes requested):

games/a.yaml
```diff
@@ -663,6 +663,22 @@
   added: '2025-11-06'
   updated: '2025-11-06'
 
+- name: Alpha Waves Classic
```
cxong, line 666:
```suggestion
- name: Alpha-Waves / Continuum
```
I don't see "classic" mentioned

ju5: Download and unzip the [CVS Repository](https://sourceforge.net/p/alpha-waves/code/) alpha-waves.zip file. Use the command `cvs -d /full/path/to/alpha-waves checkout alpha`. You will see this name in the generated source code.

cxong (approved):

PR #3748, opened 2025-11-08
Allow filtering by multiple tags at the same time

lostmsu:
- when you click on a tag filtering by it is toggled
- only games that match all the selected tags are displayed
- when no tags are selected, all games are displayed
- selected tags are displayed above the game list


cxong:
Nice work! I moved the tags to the header so it stays visible

PR #3818, opened 2025-11-19
Add Bubble Train

ju5:


games/b.yaml
```diff
@@ -1755,6 +1755,33 @@
   images:
   - https://raw.githubusercontent.com/rembound/Bubble-Shooter-HTML5/master/screenshot.png
 
+- name: Bubble Train
+  originals:
+  - Puzzle Bobble
```
cxong: despite the author's claims it looks more like puzz loop to me

PR #3829, opened 2025-11-20
Add OpenDaed

ju5:


AMDmi3:
Wow, someone have actually found it.

originals/d.yaml
```diff
@@ -1,3 +1,19 @@
+- name: The Daedalus Encounter
+  names:
+  - 'Daedalus Episode 1: Nanpasen no Alien'
+  - デイドラス エピソード1：難破船のエイリアン
+  - 데드러스 인카운터
```
cxong:
```suggestion
```
Apart from English and title in its native language (if it differs significantly) we don't need to add other languages, especially if it's just a transliteration

PR #3849, opened 2025-11-23
Add Uplink

ju5:


games/u.yaml
```diff
@@ -560,6 +560,21 @@
   url: https://unvanquished.net
   feed: https://unvanquished.net/feed
 
+- name: Uplink
+  originals:
+  - Uplink
+  type: remake
+  repo: 'https://github.com/AdrUlb/Uplink'
+  development: sporadic
+  status: playable
```
cxong: where does it show the game as playable? I don't see any significant code about game logic etc

AdrUlb: I don't know where this PR got that idea from but it certainly wasn't me. It isn't playable and isn't going to be for a while.

ju5: Based on your feedback, it has been changed to unplayable.

PR #3901, opened 2025-11-27
Add Real-time Command to Conquer

ju5:


games/r.yaml
```diff
@@ -320,6 +320,25 @@
   type: remake
   updated: 2022-11-13
 
+- name: Real-time Command to Conquer
```
cxong, line 323:
```suggestion
- name: Real Time Command and Conquer
```

ju5: I think the repo page uses an incorrect title. Refer to [mainpage](https://web.archive.org/web/20010821023218/http://www.calcware.org/member/tnt/#:~:text=Real-time%20Command%20to%20Conquer) and the source code for the official name.

PR #3912, opened 2025-11-29
Add remc1hw

ju5:


games/r.yaml
```diff
@@ -513,6 +513,25 @@
...
 
+- name: remc1hw
+  originals:
+  - Magic Carpet
+  type: remake
+  repo: 'https://github.com/turican0/remc1hw'
+  development: halted
+  status: unplayable
+  multiplayer:
+  - LAN
```
```suggestion
```
if it is unplayable I don't think it can have multiplayer

games/r.yaml
```diff
@@ -513,6 +513,25 @@
...
+  content: commercial
+  langs:
+  - C++
+  frameworks:
+  - SDL
```
```suggestion
```
I don't see SDL referenced in the source code

PR #3918, opened 2025-11-30
Add AberMUD2

ju5:


cxong:
this game was always open source so it should not be added

PR #3929, opened 2025-12-01
Add Catacomb

ju5:


originals/c.yaml
cxong, line 1: I don't think the games should be split, catacomb ii is basically the same game but with more levels

PR #3934, opened 2025-12-02
Add Baldur's Gate: Dark Alliance I & II

RoseIlanga:


originals/b.yaml
```diff
@@ -75,6 +75,60 @@
...
+  - PlayStation 4
+  - PlayStation 5
+  - Windows
+  - Xbox
+  - Xbox One
+  - Xbox Series S/X
+  meta:
+    genres:
+    - Action
+    - Adventure
```
```suggestion
```

originals/b.yaml
```diff
@@ -75,6 +75,60 @@
...
```
cxong, line 99:
```suggestion
```

RoseIlanga: Yes, #3941. I didn't know if I was supposed to make a separate pull request for everything or not.

PR #3979, opened 2025-12-11
Add Super Mario 64 (DualShock Version)

ju5:


malucard:
I'm not familiar with this repo, but this is a port, does it really count?

PR #4006, opened 2025-12-15
Add World in Conflict Massgate

ju5:


games/w.yaml
```diff
@@ -1112,6 +1112,27 @@
   images:
   - https://user-images.githubusercontent.com/3178471/150548930-9dab1e11-2997-48da-af33-6e3386017a50.gif
 
+- name: World in Conflict Massgate
+  originals:
+  - World in Conflict
+  type: official
```
```suggestion
  type: tool
```
I think this should be considered a tool because it only manages one aspect of the multiplayer function of the game, and does not contain any game logic itself. This sets it apart from MMO servers which are necessary to play the game

PR #4010, opened 2025-12-16
Add Xargon

ju5:


games/x.yaml
```diff
@@ -157,6 +157,20 @@
   images:
   - 'https://repo.openpandora.org/files/pnd/xargon_ptitseb/preview3.png'
 
+- name: Xargon (1994)
```
```suggestion
- name: Xargon
```

ju5: Conflicts with an existing entry name.

cxong: since this is the official one, the existing entry should be renamed (like Xargon (SDL))

PR #4040, opened 2025-12-21
Add Blue Cosmos

ju5:
Reference https://code.google.com/archive/p/bluecosmos/


cxong:
this shouldn't be added, it is in such an early stage it is impossible to say whether this is actually related to ikaruga or not - the description in the project seems more like a statement of intent than fact

PR #4047, opened 2025-12-22
Add Doom83

ju5:


games/d.yaml
```diff
@@ -1538,6 +1538,25 @@
   updated: 2022-07-02
   url: https://doom64ex.wordpress.com/
 
+- name: Doom83
+  originals:
+  - Doom
```
```suggestion
  - Wolfenstein 3D
```
the graphics look nothing like doom

games/d.yaml
```diff
@@ -1538,6 +1538,25 @@
...
+  type: remake
```
```suggestion
  type: similar
```

PR #4048, opened 2025-12-23
Add DROID

ju5:


cxong:
I don't think this should be added since there are a lot of robot programming games like robotwar and this one doesn't even mention robotwar

PR #4065, opened 2025-12-26
Add Pybots

ju5:


games/p.yaml
```diff
@@ -2522,6 +2522,20 @@
   added: 2025-11-08
   updated: 2025-11-08
 
+- name: Pybots
+  originals:
+  - RobotWar
```
cxong:
```suggestion
  - CROBOTS
```

ju5: CROBOTS is an open-source clone of RobotWar that is not suitable as the original game.

cxong: CROBOTS was initially released as shareware, open sourced later, so it can be an original game

PR #4099, opened 2025-12-31
Add Pang for Linux!

ju5:


originals/b.yaml
```diff
@@ -1093,17 +1093,26 @@
...
+  - Amstrad CPC
+  - Arcade
+  - Atari
+  - Commodore 64
+  - Game Boy
+  - TurboGrafx-16
+  - ZX Spectrum
+  meta:
+    genres:
+    - Platform
```
```suggestion
    - Action
    - Arcade
```
I don't think it should be considered a platform game, since many levels don't even have platforms, the player can't jump, and most of the focus is on shooting, so the platforming parts are very minor

PR #4104, opened 2026-01-01
Add NetHack

ju5:


cxong:
I don't think nethack should be added - it was always developed in the open. We can keep the existing entries though

PR #4130, opened 2026-01-05
Add retruxx

mindflower:


games/r.yaml
```diff
@@ -791,6 +791,25 @@
...
+  - Hard Truck Apocalypse
+  type: remake
+  repo: 'https://github.com/mindflower/retruxx'
+  development: active
+  status: semi-playable
+  content: commercial
+  langs:
+  - C++
+  - C
+  - Lua
```
```suggestion
```
no lua in source code

mindflower: embedded lua is used there for writing game scripts. is it better to move lua to frameworks or remove it from the description?

mindflower: fixed

PR #4143, opened 2026-01-06
Add Campo Minado

ju5:


games/c.yaml
```diff
@@ -193,6 +193,29 @@
   images:
   - 'https://www.calculatorti.com/media/images/programs/nethack-rogue.jpg'
 
+- name: Campo Minado
+  originals:
+  - Minefield
```
```suggestion
  - Mined-Out
```
https://en.wikipedia.org/wiki/Mined-Out

games/c.yaml
```diff
@@ -193,6 +193,29 @@
...
+  type: similar
```
```suggestion
  type: clone
```

PR #4167, opened 2026-01-09
Add MasterPiece Assembly

ju5:


games/m.yaml
```diff
@@ -454,7 +454,7 @@
 
 - name: MasterPiece
   originals:
-  - Columns
+  - Acid Drop
```
cxong, line 457: why is this changed to acid drop? It mentions and looks like columns

ju5: The introduction to the C++ version of MasterPiece mentions Acid Drop:
https://lostsidedead.biz/game/MasterPiece3D/#:~:text=Acid%20Drop
All three MasterPiece games were developed by the same creator.

PR #4171, opened 2026-01-10
Add Minefield

ju5:


games/c.yaml
```diff
@@ -207,6 +207,28 @@
   images:
   - 'https://www.calculatorti.com/media/images/programs/nethack-rogue.jpg'
 
+- name: Campo de Minas
```
```suggestion
- name: Minefield
```

ju5: There are too many games called Minefield; I think it's better to use the alias mentioned in its url.

cxong: the alias looks like a Spanish translation and the name should be Minefield, which the source article and code both mention

PR #4206, opened 2026-01-16
Add DOOM-FX

ju5:


games/d.yaml
```diff
@@ -1573,6 +1573,25 @@
...
+  repo: 'https://github.com/RandalLinden/DOOM-FX'
+  url: >-
+    https://www.pcgamer.com/source-code-for-the-snes-version-of-doom-has-been-released/
+  development: complete
+  status: playable
+  content: commercial
+  langs:
+  - Assembly
+  frameworks:
+  - Reality
```
```suggestion
```
this is a one off engine for this game only so no need to add it

PR #4243, opened 2026-01-20
Add Freefall Source Code

ju5:


originals/f.yaml
```diff
@@ -340,6 +340,17 @@
   platforms:
   - Windows
 
+- name: Fishfall
```
cxong:
```suggestion
- name: Free Fall (Color Dreams)
```

originals/f.yaml
```diff
@@ -340,9 +340,9 @@
   platforms:
   - Windows
 
-- name: Fishfall
+- name: Free Fall (Color Dreams)
```
cxong: need to reorder this

PR #4251, opened 2026-01-21
Add Reconstructed Commander Keen 1-3

ju5:


games/r.yaml
```diff
@@ -458,6 +458,22 @@
   added: 2026-01-12
   updated: 2026-01-12
 
+- name: Reconstructed Commander Keen 1-3
+  originals:
+  - Commander Keen Series
+  type: remake
+  url: 'https://pckf.com/viewtopic.php?t=18248'
+  development: active
+  status: playable
```
cxong, line 467:
```suggestion
  status: halted
```
although it's recent it is posted in a forum so there's no indication it will be actively developed from now on

PR #4266, opened 2026-01-24
Add No One Lives Forever Modernizer

ju5:


games/n.yaml
```diff
@@ -775,6 +775,32 @@
   video:
     youtube: Y3ugMLPDYQ8
 
+- name: No One Lives Forever Modernizer
+  originals:
+  - 'The Operative: No One Lives Forever'
+  type: remake
```
cxong:
```suggestion
  type: tool
```

games/n.yaml
```diff
@@ -775,6 +775,32 @@
...
+  repo: 'https://github.com/haekb/nolf1-modernizer'
+  url: 'https://haekb.itch.io/nolf-modernizer'
+  development: halted
+  status: playable
+  multiplayer:
+  - Online
+  - LAN
```
cxong:
```suggestion
```
tools don't need multiplayer - it would just be the same as the original game

games/n.yaml
```diff
@@ -775,6 +775,32 @@
...
+  langs:
+  - C++
+  frameworks:
+  - LithTech
+  - DirectX
```
cxong:
```suggestion
  - SDL2
```

PR #4267, opened 2026-01-24
Add lithtech

ju5:


games/s.yaml
```diff
@@ -1917,6 +1917,28 @@
   added: 2018-06-15
   updated: 2023-10-27
 
+- name: 'Shogo: Mobile Armor Division'
+  originals:
+  - 'Shogo: Mobile Armor Division'
+  type: official
+  repo: 'https://github.com/haekb/ltel-shogo'
+  url: 'https://www.golem.de/9811/2181.html'
```
```suggestion
```
this url is unrelated to the repo

games/s.yaml
```diff
@@ -1917,6 +1917,28 @@
   added: 2018-06-15
   updated: 2023-10-27
 
+- name: 'Shogo: Mobile Armor Division'
```
cxong: adding this game in this form is misleading as it's not the full game source. It is only the gameplay source and doesn't include engine code, so will not create a working game. It is useful for limited purposes like making mods. So it's more like an SDK. At the very least the type needs to be tool, and the name needs to make clear what it is. Having the name match the game name is misleading

games/s.yaml
```diff
@@ -1917,6 +1917,28 @@
...
+  originals:
+  - 'Shogo: Mobile Armor Division'
+  type: official
+  repo: 'https://github.com/haekb/ltel-shogo'
```
```suggestion
  repo: https://www.bluesnews.com/files/shogo/tools/shogo-src.shtml
```
while the repo is pretty much the source code unaltered, the project implies it will be part of a different remake project, so would be unsuitable for representing the source code release itself

ju5: The download link is no longer valid.

ju5:
Replace with Shogo's engine lithtech.

Katana-Steel:
Lithtech is really the Monolith's LithTech Jupiter **game engine** and really only NOLF2 is the only playable instantiation.

there are a few screenshots of NOLF2 in the repo, but no videos as far as I know of.

PR #4268, opened 2026-01-24
Add DGEM

ju5:
Tip: Use [gfalist](https://github.com/mmuman/gfalist) to view the source code within GFA files.


games/d.yaml
```diff
@@ -796,6 +796,23 @@
...
+  - Dungeon Master
+  type: clone
+  url: 'https://ptonthat.fr/dgem/'
+  development: halted
+  status: playable
+  content: commercial
+  langs:
+  - BASIC
+  licenses:
+  - As-is
```
```suggestion
  - Custom
```

PR #4286, opened 2026-01-28
Add Hashlife

ju5:


cxong:
I don't think we should add game of life because it's not a traditional video game, it is a well known simulation that has multiple implementations, and there is no original game

PR #4291, opened 2026-01-28
Add The Almost of ZZT

ju5:


asiekierka:
The Reconstruction of ZZT is already on the page, and I would prefer that to remain the "canonical" version of open source ZZT due to its clearer provenance and being based on a newer version of ZZT. The Almost of ZZT is meant solely as a historical artifact, not as a playable or forkable entity.

PR #4292, opened 2026-01-29
Add Breakout (Ken Silverman)

ju5:


games/w.yaml
```diff
@@ -519,6 +519,24 @@
   - https://www.golem.de/1302/97509-53272-webquake_rc.jpg
   - https://images.kartones.net/posts/kartonesblog/webquake.webp
 
+- name: Welcome to Breakout!
```
```suggestion
- name: Breakout (Ken Silverman)
```
the only place where this is mentioned is in the title splash - Ken Silverman is way more notable than this game name

PR #4347, opened 2026-02-07
Add Scroller

ju5:


cxong:
I don't think this should be added as I can't find evidence that this was released publicly before the source code release

ju5:
@cxong According to https://www.mobygames.com/game/35807/scroller/, Scroller was released prior to Super Silverbrothers (1991).
Comparing the two snapshots reveals that 10/28/2004 is actually the source code release date:
https://web.archive.org/web/20041022080711/https://advsys.net/ken/download.htm#scroller
https://web.archive.org/web/20050206213239/https://advsys.net/ken/download.htm#scroller

Early snapshots of SCROLLER.ZIP did not include source code:
https://web.archive.org/web/20010501000000*/http://advsys.net/ken/scroller.zip

PR #4348, opened 2026-02-07
Add Hover Strike

ju5:


cxong:
there is already an official hover strike entry

ju5:
@cxong The new addition is the original version, while the existing one is the remastered version, Hover Strike: Unconquered Lands.

PR #4366, opened 2026-02-10
Add C64 Wizard of Wor Disassembly

ju5:


dabadab:
Hello!

I'm not familiar with your project but my WoW disassembly is certainly NOT a remake / clone - it's the original one. When you compile it you get exactly the original game.

ju5:
It's not the official source code; we're treating it as a remake.

games/c.yaml
```diff
@@ -121,6 +121,24 @@
...
+  type: remake
+  repo: 'https://github.com/dabadab/wizardofwor'
+  development: halted
+  status: playable
+  multiplayer:
+  - Competitive
+  - Local
+  - Co-op
+  langs:
+  - C++
```
```suggestion
  - Assembly
```

PR #4377, opened 2026-02-11
Add Army Men 1998 Source Code and Tools

ju5:


games/a.yaml
```diff
@@ -2003,6 +2003,22 @@
...
+  type: official
+  repo: 'https://github.com/OnyeNacho/ArmyMen'
+  url: 'https://archive.org/details/armymen_202209'
+  development: complete
+  status: playable
+  content: free
+  langs:
+  - C++
+  licenses:
+  - CC0
```
```suggestion
  - As-is
```
I don't see an explicit copyright grant by the original authors and it is unclear what providence the code has

PR #4389, opened 2026-02-15
Add three-descent

ju5:


mrdoob:
You may want to add https://mrdoob.github.io/three-quake/ (https://github.com/mrdoob/three-quake) too.

PR #4390, opened 2026-02-15
Add c4p

ju5:


cxong:
I think we should disallow clones of board games as per https://github.com/opengaming/osgameclones/discussions/2789 (unless the game has an official version to clone)

PR #4408, opened 2026-02-21
Add Final Fantasy disassemblies

ju5:


everything8215:
Hi @ju5, I’d like to request that you not add these repos to the Open Gaming website. These are disassemblies of the original assembly code, not “remakes” as defined by this site. My main concern is with copyright implications from the developer. These repos are already listed on the RetroReversing website which is more appropriate. If you would like to add them under a “Disassembly” or “Reverse Engineering” category that could be acceptable, though I’m not sure if that is the intent of Open Gaming.

ju5:
@everything8215 OSGC does include reverse-engineered projects, and they are categorized as "remakes". If the original code is not open-source, it will be marked in red on the page.
Either way, I think we can wait for the maintainer to decide whether these repos should be added.

originals/f.yaml
```diff
@@ -269,6 +269,48 @@
...
+    - Fantasy
+
+- name: Final Fantasy IV
+  names:
+  - ファイナルファンタジーIV
+  external:
+    website: 'https://www.jp.square-enix.com/game/detail/ff4/'
+    wikipedia: Final Fantasy IV
+  platforms:
+  - NES
```
cxong:
```suggestion
  - SNES
```

originals/f.yaml
```diff
@@ -286,6 +328,34 @@
...
     - Fantasy
 
+- name: Final Fantasy V
+  names:
+  - ファイナルファンタジーV
+  external:
+    website: 'https://www.jp.square-enix.com/game/detail/ff5/'
+    wikipedia: Final Fantasy V
+  platforms:
+  - NES
```
cxong:
```suggestion
  - SNES
```

originals/f.yaml
```diff
@@ -286,6 +328,34 @@
...
+    - Fantasy
+
+- name: Final Fantasy VI
+  names:
+  - ファイナルファンタジーVI
+  external:
+    website: 'https://www.jp.square-enix.com/game/detail/ff6/'
+    wikipedia: Final Fantasy VI
+  platforms:
+  - NES
```
cxong:
```suggestion
  - SNES
```

PR #4440, opened 2026-02-27
Add SharpMC

ju5:


games/s.yaml
```diff
@@ -1817,6 +1817,42 @@
...
+  type: remake
+  repo: 'https://github.com/SharpMC/SharpMC'
+  development: halted
+  status: semi-playable
+  multiplayer:
+  - Online
+  langs:
+  - 'C#'
+  frameworks:
+  - MonoGame
```
```suggestion
  - Mono
```
monogame is not the same as mono

PR #4470, opened 2026-03-03
Added Prey (2006) Source Port

FriskTheFallenHuman:


originals/p.yaml
```diff
@@ -1121,6 +1121,24 @@
     themes:
     - Prehistoric
 
+- name: Prey
+  names:
+  - Prey (2006)
+  - Prey A Talon Brave Game
```
cxong:
```suggestion
```
the final game wasn't released using this title so don't need to include here

PR #4472, opened 2026-03-04
Add FF1 Disassembly with Bugfix & Enhancement Patches

ju5:


games/f.yaml
```diff
@@ -265,6 +265,20 @@
...
+  originals:
+  - Final Fantasy
+  type: remake
+  url: 'https://www.romhacking.net/documents/663/'
+  development: complete
+  status: playable
+  langs:
+  - Assembly
+  licenses:
+  - As-is
```
cxong, line 278:
```suggestion
  - As-is
  content: commercial
```

PR #4516, opened 2026-03-11
Add ff1snes

ju5:


games/f.yaml
```diff
@@ -280,6 +280,22 @@
...
+  originals:
+  - Final Fantasy
+  type: remake
+  repo: 'https://github.com/Entroper/ff1snes'
+  development: sporadic
+  status: semi-playable
+  content: free
+  langs:
+  - Assembly
+  - C++
```
```suggestion
  - 'C#'
```
I recommend not trusting github language detection as it often gets it wrong

games/f.yaml
```diff
@@ -280,6 +280,22 @@
...
   updated: 2026-03-05
 
+- name: ff1snes
+  originals:
+  - Final Fantasy
+  type: remake
+  repo: 'https://github.com/Entroper/ff1snes'
+  development: sporadic
+  status: semi-playable
+  content: free
```
```suggestion
  content: commercial
```

PR #4530, opened 2026-03-13
Update Spirit of Speed 2002, XDigger, Vectoroids, Buck Rogers

ju5:


games/s.yaml
```diff
@@ -4355,7 +4355,8 @@
   originals:
   - Spirit of Speed 1937
   type: similar
-  repo: 'https://github.com/LemonHaze420/SOS2002'
+  repo: 'https://github.com/LemonHaze420/SOS2002_src'
+  url: 'https://github.com/LemonHaze420/SOS2002'
```
```suggestion
```
this URL is misleading as there is no obvious link with the source code
would suggest putting this context into the info field

games/v.yaml
```diff
@@ -202,9 +202,11 @@
   - Asteroids
   status: playable
   type: remake
-  updated: 2022-05-03
+  added: 2015-04-07
+  updated: 2026-03-13
   url: http://www.newbreedsoftware.com/vectoroids/
-  
+  repo: https://github.com/midzer/vectoroids
```
cxong, line 208:
```suggestion
  repo: https://github.com/midzer/vectoroids
  video:
    youtube: ISVTXkaviVM
```

PR #4545, opened 2026-03-15
Add TetrisNESDisasm

ju5:


originals/c.yaml
```diff
@@ -718,6 +718,16 @@
     themes:
     - Anime
 
+- name: classic Tetris
```
cxong:
```suggestion
- name: Tetris (NES)
```
"classic Tetris" is easily confused with "Tetris Classic"

PR #4568, opened 2026-03-18
Add Match3

ju5:


cxong:
there's no mention of bejeweled - not all match 3 games are clones of it

PR #4575, opened 2026-03-19
Add Gem Drop X

ju5:


games/g.yaml
```diff
@@ -387,6 +387,32 @@
...
+  development: halted
+  status: playable
+  content: free
+  langs:
+  - C
+  frameworks:
+  - SDL
+  licenses:
+  - As-is
+  info: Atari 8-bit port of Gem Drop
```
cxong: this is a port FROM the atari version

PR #4580, opened 2026-03-20
Add Pikmin

ju5:


originals/p.yaml
```diff
@@ -478,6 +478,24 @@
     genres:
     - Puzzle
 
+- name: Pikmin
+  names:
+  - 'New Play Control: Pikmin'
```
cxong:
```suggestion
```
this name already includes pikmin so doesn't help much

PR #4588, opened 2026-03-21
Add Mr.Bomber game.

chemodax:
Add Mr.Bomber game:
https://mrboom.app/

### DIFF
--- a/games/m.yaml
+++ b/games/m.yaml
@@ -3230,6 +3230,32 @@
   images:
   - 'https://www.ticalc.org/archives/files/ss/705/70590.gif'
 
+- name: Mr.Bomber
+  originals:
+  - Bomberman
+  type: remake
+  repo: 'https://github.com/rinrab/mr-bomber'
+  url: 'https://mrboom.app/'
+  development: sporadic
+  status: playable
+  multiplayer:
+  - Competitive
+  - Hotseat
+  content: open
+  langs:
+  - 'C#'
+  frameworks:
+  - MonoGame
+  licenses:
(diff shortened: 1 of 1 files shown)


games/m.yaml
```diff
@@ -3230,6 +3230,32 @@
...
+  - Bomberman
+  type: remake
+  repo: 'https://github.com/rinrab/mr-bomber'
+  url: 'https://mrboom.app/'
+  development: sporadic
+  status: playable
+  multiplayer:
+  - Competitive
+  - Hotseat
+  content: open
```
cxong, line 3244:
```suggestion
  content: free
```
mr boom assets are freeware not open

chemodax: Well, there is repository with original assembly source code and assets. It lists them as GPL:
https://github.com/reymontero/mrboom30

rinrab: Perhaps the relevant repository that the assets belong to is actually this one https://github.com/Javanaise/mrboom-libretro.

Which is published under MIT license, not GPL.

I couldn't tell for sure.

cxong: The libretro port is from the original author Remdy Software so MIT is fine. I don't know about GPL though, since that repo is not from the original author and I can't see where it was released as GPL in the past
As for the assets it looks like they aren't exact copies, only that the character designs are extremely similar to games like super bomberman. It might infringe on trademark but at least for copyright it should be ok, so resolving this convo

rinrab:
Thanks for adding the game!

PR #4618, opened 2026-03-26
Add Source Code For RBI 94 Sega Genesis

ju5:


originals/r.yaml
```diff
@@ -8,6 +8,17 @@
     genres:
     - Racing
 
+- name: "R.B.I. Baseball '94"
```
cxong: this should be sorted before R.C

PR #4620, opened 2026-03-26
Add foodfight-7800

ju5:


games/f.yaml
```diff
@@ -977,6 +977,25 @@
   type: remake
   updated: 2022-06-14
 
+- name: foodfight-7800
```
cxong, line 967: sort this before football manager

originals/f.yaml
```diff
@@ -598,6 +598,16 @@
...
 
+- name: Food Fight
+  external:
+    website: 'https://www.mobygames.com/game/11242/food-fight/'
+    wikipedia: Food Fight (video game)
+  platforms:
+  - Arcade
+  - Atari
+  - Windows
+  - Xbox 360
```
cxong, line 609:
```suggestion
  - Xbox 360
  meta:
    genres:
    - Action
```

PR #4642, opened 2026-03-31
Add Invaders

ju5:


games/i.yaml
```diff
@@ -357,6 +357,22 @@
   - >-
     https://camo.githubusercontent.com/a6ea5fa0c684804202cce2538a73c0e915b81f5e0079c59a5bedea1c56fb4751/687474703a2f2f692e696d6775722e636f6d2f713459355661652e676966
 
+- name: Invaders
```
cxong: there's an existing game with the same name but different capitalization, we should ensure game names are unique regardless

PR #4682, opened 2026-04-10
Add Attack of the Polps!

ju5:


cxong:
not really a clone of space invaders, only mention is the category which is quite broad

PR #4698, opened 2026-04-13
Add Exile disassembly

ju5:


tom-seddon:
I don't think my exile-disassembly project is suitable for this repo, as it's not FOSS and it's not a remake. It might make sense by way of link to the original if somebody does a FOSS remake.

Rather than link to my repo, I think a better link would be to the original disassembly, from which mine was mechanically derived: https://level7.org.uk/miscellany/exile-standard-disassembly.txt

Mine did include some stuff previously not covered by the original, but there's an updated disassembly including those parts now: https://level7.org.uk/miscellany/exile-enhanced-disassembly.txt

ju5:
OSGC classifies disassemblies as remakes and highlights projects without an open-source licence in red.

PR #4704, opened 2026-04-15
Add FFED3D

ju5:
Reference https://code.google.com/archive/p/ffed3d/

### DIFF
--- a/games/f.yaml
+++ b/games/f.yaml
@@ -331,6 +331,33 @@
   added: 2026-02-11
   updated: 2026-03-03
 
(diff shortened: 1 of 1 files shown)


games/f.yaml
```diff
@@ -331,6 +331,33 @@
...
+  development: halted
+  status: playable
+  content: commercial
+  langs:
+  - Assembly
+  - C++
+  frameworks:
+  - DirectX
+  licenses:
+  - LGPL3
```
```suggestion
  - As-is
```
no license in source or repo

PR #4720, opened 2026-04-19
Add Bu!4096

ju5:


cxong:
I think the "Space Invaders" category is not well named at allegro.cc, only a few games are actually clones of space invaders, the others seem to be fixed shooters only which is a subgrenre

PR #4731, opened 2026-04-21
Add Aidyn Chronicles Decompilation

ju5:


games/a.yaml
```diff
@@ -467,6 +467,21 @@
   added: 2025-11-24
   updated: 2025-11-24
 
+- name: Aidyn Chronicles Decompilation
+  originals:
+  - 'Aidyn Chronicles: The First Mage'
+  type: remake
+  repo: 'https://github.com/blackgamma7/Aidyn'
+  development: active
+  status: playable
```
```suggestion
  status: unplayable
```

blackgamma7: also technically a decomp, not a remake.

PR #4735, opened 2026-04-22
Add mb-decomp

ju5:


games/m.yaml
```diff
@@ -857,6 +857,22 @@
   - https://cdn-ak.f.st-hatena.com/images/fotolife/H/Hiesuke/20240419/20240419153401.png
   - https://cdn-ak.f.st-hatena.com/images/fotolife/H/Hiesuke/20240522/20240522232117.gif
 
+- name: mb-decomp
+  originals:
+  - Monkey Ball
```
```suggestion
  - Super Monkey Ball
```
same game, arcade title

ju5: The Super Monkey Ball entry is linked to the entire series, whereas mb-decomp only applies to the first game in the series, Monkey Ball.

cxong: this is fine because the gamecube game is a port of the arcade game and thus the same game. We don't want separate entries for every platform port

games/m.yaml
```diff
@@ -857,6 +857,22 @@
   - https://cdn-ak.f.st-hatena.com/images/fotolife/H/Hiesuke/20240419/20240419153401.png
   - https://cdn-ak.f.st-hatena.com/images/fotolife/H/Hiesuke/20240522/20240522232117.gif
 
+- name: mb-decomp
+  originals:
+  - Super Monkey Ball
+  type: remake
+  repo: 'https://github.com/Yoshimaster96/mb-decomp'
+  development: halted
+  status: unplayable
```
cxong, line 866:
```suggestion
  status: unplayable
  platforms:
  - Arcade
```

PR #4739, opened 2026-04-22
Add superstuntcycle-7800

ju5:


originals/m.yaml
```diff
@@ -1479,6 +1479,19 @@
   - Amstrad CPC
   - DOS
 
+- name: MotorPsycho
```
cxong:
```suggestion
- name: Super Stunt Cycle
```
it appears these are two different games but the creator confusingly reused the name at some point

PR #4756, opened 2026-04-28
Add Shadow Warrior for Linux

ju5:


icculus:
...it's not a clone, though...?  :)

ju5:
Port is also acceptable.

PR #4771, opened 2026-04-30
Add vkHexen

ju5:


atsb:
Don't add this, vkHexen while compilable, does not have a functioning rendering system.

PR #4783, opened 2026-05-01
Add XPilot

ju5:


cxong:
xpilot doesn't really belong because it was always open source. I think we should remove it entirely

PR #4807, opened 2026-05-05
Add Amidar500

ju5:


games/a.yaml
```diff
@@ -1147,6 +1147,28 @@
...
   - 'https://keasigmadelta.com/wp-content/uploads/2025/01/AmiCraftNova-1.jpg'
 
+- name: Amidar500
+  originals:
+  - Amidar
+  type: remake
+  repo: 'https://github.com/jotd666/amidar500'
+  url: 'https://jotd666.itch.io/amidar500'
+  development: active
+  status: semi-playable
```
```suggestion
  status: playable
```

games/a.yaml
```diff
@@ -1147,6 +1147,28 @@
...
+  multiplayer:
+  - Local
```
```suggestion
```
how is this multiplayer?

ju5: I'm guessing the remake will replicate the original's multiplayer mode, but I'm not sure what type it will be: https://en.wikipedia.org/wiki/Amidar

cxong: https://www.mobygames.com/game/11418/amidar/ claims it's one player. I can't find a screenshot or video online showing multiplayer, so if you are able to find one please put in the correct type of multiplayer

jotd666:
it's a 2-player game but on turns. Players are switched when the current
player loses a life, like Pacman, all other similar games or pinball
machines! Note that the amiga port doesn't have 2p mode as it's not a
transcode of the arcade game. One player only.

Le ven. 29 mai 2026 à 11:26, congusbongus ***@***.***> a
écrit :

PR #4808, opened 2026-05-06
Added Wizball

craigchandler:
Added Wizball remake

### DIFF
--- a/games/w.yaml
+++ b/games/w.yaml
@@ -1001,7 +1001,7 @@
   - C++
   frameworks:
   - SDL2
-  - GLES2
+  - OpenGL
   licenses:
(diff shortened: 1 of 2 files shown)


craigchandler:
Should I drop GLES2 or leave it there? (Only uses ES subset of OpenGL on portmaster devices)

games/w.yaml
```diff
@@ -985,6 +985,34 @@
...
+  status: playable
+  multiplayer:
+  - Local
+  - Co-op
+  content: open
+  langs:
+  - C++
+  frameworks:
+  - SDL2
+  - GLES2
```
```suggestion
  - OpenGL
```

PR #4825, opened 2026-05-12
Add Super-Mario-64-DS-Remastered

ju5:


originals/s.yaml
```diff
@@ -2681,6 +2681,8 @@
     - Platform
 
 - name: Super Mario 64
+  names:
+  - Super Mario 64 DS
```
cxong:
```suggestion
```
just a port/rerelease name, don't need it

games/s.yaml
```diff
@@ -6095,6 +6095,26 @@
   added: 2026-01-23
   updated: 2026-01-23
 
+- name: Super-Mario-64-DS-Remastered
+  originals:
+  - Super Mario 64
+  type: remake
```
```suggestion
  type: tool
```
looks like this is a mod that requires some other game to play

PR #4845, opened 2026-05-17
Add Winblockade

ju5:


originals/b.yaml
```diff
@@ -918,6 +918,17 @@
     genres:
     - Arcade
 
+- name: Blockade
+  external:
+    website: 'https://www.mobygames.com/game/80291/blockade/'
```
cxong: should be https://www.mobygames.com/game/68275/blockade/ instead

PR #4870, opened 2026-05-25
Add Conker's Bad Fur Day Decompilation

ju5:


games/c.yaml
```diff
@@ -2774,6 +2774,24 @@
   images:
   - 'https://elite.bbcelite.com/images/general/Elite-Commodore64.png'
 
+- name: Conker's Bad Fur Day Decompilation
+  originals:
+  - Conker's Bad Fur Day
+  type: remake
+  repo: 'https://github.com/mkst/conker'
+  url: 'https://conker.deco.mp/'
+  development: active
```
mkst: not really..

ju5: Could you be more specific?

mkst: Who's actively developing it?

ju5: A project is considered active if it has had commits within the last month.

mkst: That was to mitigate the `pull_request_target` ci workflow. Nothing meaningful has been done in terms of decomp for a long time.

PR #4875, opened 2026-05-25
Add Paciana Dones

ju5:


games/p.yaml
```diff
@@ -235,6 +235,32 @@
...
+  originals:
+  - Pac-Man
+  - Sokoban
+  type: similar
+  url: 'https://www.allegro.cc/depot/PacianaDones'
+  development: halted
+  status: playable
+  platforms:
+  - DOS
+  - Linux
```
cxong, line 248:
```suggestion
```

ju5: The readme.txt in the zip mentions these three platforms.

PR #4884, opened 2026-05-28
Add Gravitar

ju5:


jotd666:
Gravitar doesn't run properly. Code has been ported from 6502 to 68000 fairly accurately but the line drawing routines are broken. Game displays highscores then starts attract mode and crashes because of wrong linedraw.

PR #4904, opened 2026-06-07
Add Silent Hill Decompilations

ju5:


games/s.yaml
```diff
@@ -2411,6 +2411,24 @@
...
+  type: remake
+  repo: 'https://github.com/Vatuu/silent-hill-decomp'
+  url: 'https://decomp.dev/Vatuu/silent-hill-decomp'
+  development: very active
+  status: semi-playable
+  platforms:
+  - PlayStation
+  content: commercial
+  langs:
+  - C++
```
dreamingmoths: thanks for adding the sh1 & sh2+3 repos!

note that sh1 also uses c, not c++. unless you mean the port, but that's not linked here

PR #4947, opened 2026-07-02
Add hugo.js

ju5:


originals/h.yaml
```diff
@@ -727,6 +727,18 @@
   - DOS
   - Windows
 
+- name: 'Hugo: På Nye Eventyr'
+  external:
+    website: 'https://www.mobygames.com/game/259829/hugo-pa-nye-eventyr/'
+  platforms:
+  - Amiga
+  - DOS
```
cxong, line 735:
```suggestion
```

originals/h.yaml
```diff
@@ -727,6 +727,18 @@
...
+  meta:
+    genres:
+    - Adventure
```
cxong, line 739:
```suggestion
    - Action
```

originals/h.yaml
ju5: It's also a DOS adventure game, see https://www.igdb.com/games/hugo-pa-nye-eventyr

PR #4978, opened 2026-07-14
Add Harrier Attack disassembly

ju5:


pobtastic:
Technically... the language is "Assembly"...  Although I appreciate that it's written using Skoolkit, and the assembly is generated from that.

ju5:
The project itself doesn't include Assembly, so I think “Skool” makes more sense.

pobtastic:
That’s fine, there isn’t a skool language though - it’s a skoolkit control file (mainly) which generates an .asm file (and the HTML outputs for the web write-ups).

originals/h.yaml
```diff
@@ -182,6 +182,18 @@
...
+  external:
+    website: 'https://www.mobygames.com/game/14998/harrier-attack/'
+    wikipedia: Harrier Attack
+  platforms:
+  - Amstrad CPC
+  - Commodore 64
+  - ZX Spectrum
+  meta:
+    genres:
+    - Flight
```
cxong:
```suggestion
    - Shmup
```

PR #4984, opened 2026-07-22
Add Starfox 64: Recompiled

ju5:


worleydl:
My fork isn't made for general use as it contains modifications tailored specifically to UWP environments.  The actual upstream account (`sonicdcer`) has been seemingly shadowbanned which may cause difficulties producing a build because the decomp utilities were provided there too.

PR #4985, opened 2026-07-25
Update: Freeserf, freeVikings, Frozen Bubble, Galaxy Forces V2, Hypercycles

ju5:


games/f.yaml
```diff
@@ -2516,8 +2518,11 @@
   platforms:
   - None
   repo: https://sourceforge.net/projects/freevikings/
+  url: http://freevikings.wz.cz/
```
```suggestion
```
site doesn't work in http

ju5: Switched to the archive page.

PR #4990, opened 2026-07-29
Add Contra US Annotated Disassembly

ju5:


vermiceli:
To be clear https://github.com/vermiceli/nes-contra-us isn't a remake of Contra. It's just an annotated disassembly.  It does not contain any graphics nor encoded audio.  Perhaps there is another, more appropriate `type` that can be used.

PR #3944, opened 2025-12-02
Add many games

M0REKZ:
Close #3766 
Close #3875

PR #5041, opened 2026-08-29
Add mighy mike

M0REKZ:
was part of #3944

### DIFF
--- a/games/m.yaml
+++ b/games/m.yaml
@@ -2347,6 +2347,10 @@
   url: 'https://jorio.itch.io/mightymike'
   development: sporadic
   status: playable
+  platforms:
+  - Linux
+  - OSX
+  - Windows
   content: free
   langs:
   - C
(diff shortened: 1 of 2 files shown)


originals/m.yaml
```diff
@@ -1206,6 +1206,19 @@
   - Might and Magic 8 Day of the Destroyer
   - Might and Magic 8
 
+- name: Mighty Mike
+  names:
+  - Power Pete
+  external:
+    website: 'https://pangeasoft.net/mightymike/'
```
cxong, line 1213:
```suggestion
    website: 'https://pangeasoft.net/mightymike/'
    wikipedia: Power Pete
```

games/m.yaml
```diff
@@ -2339,6 +2339,36 @@
...
+  development: sporadic
+  status: playable
+  platforms:
+  - Linux
+  - OSX
+  - Windows
+  content: free
+  langs:
+  - C
+  - C++
```
```suggestion
```

M0REKZ: There is 1 cpp file in the repo https://github.com/jorio/MightyMike/blob/master/src/Boot.cpp

cxong: yeah it's very minor so I think it doesn't count

Issue #5016, opened 2026-08-14
OpenSyobonAction is probably not GPL


M0REKZ:
https://osgameclones.com/shobon-no-action

Please see M0REKZ/SyobonKZ#2 and linked issues

M0REKZ:
Probably should be marked as "as-is" or "custom" instead

PR #5044, opened 2026-08-29
Add SyobonKZ

M0REKZ:
probably need to take a look into #5016 first, here i specified the license as 'Multiple'

### DIFF
--- a/games/s.yaml
+++ b/games/s.yaml
@@ -8899,6 +8899,37 @@
   video:
     vimeo: 8942527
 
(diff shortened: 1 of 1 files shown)


M0REKZ:
Changed it back to GPLv2 (https://github.com/M0REKZ/SyobonKZ/issues/2#issuecomment-5370107757) I'm also probably the only one in 2026 actually overcomplicating about the open syobon action thing (it's been years xd...)

Issue #5046, opened 2026-08-29
Add SMBX Branches


M0REKZ:
#1980

> +KZ:
> There is also some other forks/clones for this game that must be added, but for now i would only link these.

This list has many of the main ones, not all are open source though

https://wohlsoft.ru/pgewiki/Branches_of_Super_Mario_Bros._X

M0REKZ:
for SMBX-38A (unplayable decomp): https://wohlsoft.ru/forum/viewtopic.php?f=58&t=4457

PR #5048, opened 2026-08-30
Add some SMBX Branches

M0REKZ:
opengaming/osgameclones#5046

Also updated TheXTech info


games/t.yaml
```diff
@@ -1116,7 +1116,7 @@
   - DS
   - iOS
   - Linux
-  - Nintendo Switch
```
Wohlstand: You forgot to add the Haiku!

games/t.yaml
```diff
@@ -1116,7 +1116,7 @@
   - DS
```
Wohlstand: Not the plain DS, but DSi, the plain DS has only 4 MB RAM, it's too small, the DSi offers 16 MB RAM, and therefore it works as the weakest supported target.

games/t.yaml
M0REKZ: The page does not have Haiku and DSi in the platform list D: https://osgameclones.com/add_game.html

Wohlstand: Bummer, however, what if add that to that page?

Wohlstand: I checked briefly, and seems to add new platforms it's need to fill this thing: https://github.com/opengaming/osgameclones/blob/master/schema/games.yaml#L175
And ye, the OSX is now macOS, need to rename eventually, but that's not your task.

cxong: for now I'd like to keep BeOS and Haiku under the same entry

games/t.yaml
M0REKZ: Removed DS for now to avoid confusion https://github.com/opengaming/osgameclones/pull/5048/commits/eb1a45916ffe032f5e75b641ae488b2008e78df1

games/t.yaml
M0REKZ: Not sure if BeOS can run newer Haiku software, to avoid confusion I would leave it without BeOS

Wohlstand: No, it can't. The original BeOS used the GCC 2 as a base, newer Haiku software built with GCC 7 and newer, the libc is completely incompatible. Technically it's possible if build a kind of a transition wrapper and inject newer libs into BeOS, but, build the game for 32bit architecture since Haiku build uses mostly 64bit environment without backward compatibility to old BeOS software. 32bit Haiku has a combo of GCC 2 and GCC 7 to choose what the type of libc should be used to run the app, or the compatibility mode, or the modern.

PR #5052, opened 2026-09-02
Add Ellaz: eight open-source browser clones (2048, Snake, Minesweeper, Tetris, Simon, Bejeweled, Picross, Wordle)

ytrofr:
Adds eight games from [Ellaz](https://ellaz.fun), an MIT-licensed, ad-free family games site (source: https://github.com/Sigmafier/ellaz). Each entry links the playable page and the shared repository; all eight are `Web`, TypeScript, playable, under active development.

`type: similar` where the mechanic diverges from the original or the original is a trademark (Bejeweled, Picross Series); `type: clone` for the rest.

`ai: true` on all eight, since the project contains LLM-generated code.

`images` on two of the eight, and deliberately not on the other six. Our own share cards
are content-hashed, so linking one would 404 the next time the content changes - that is
why the first version of this PR listed none at all. 2048 and Snake are also published on
itch.io, whose cover URLs are stable while the cover stands, so those two now carry one.
The remaining six have no hosted image anywhere yet; they stay imageless rather than
pointing at a URL that will rot.

The commit also registers `Preact` in `.github/workflows/pr_check.py`. The entries already
named the framework that actually ships - this project aliases `react`/`react-dom` onto
`preact/compat` - and the vocabulary simply lacked the word. Happy to split that line into
its own PR if you would rather keep this one to game entries.

Validated against `schema/games.yaml` fetched today - required keys, every enum, platforms, licenses, both dates - and every `originals` name was checked to exist in `originals/`. Entries are inserted in alphabetical position between `EliteGL` and `ElonaFoobar`.


games/e.yaml
```diff
@@ -927,6 +927,186 @@
...
+  - MIT
+  ai: true
+  info: Repeat-the-sequence memory game with four coloured pads and tones, built for young children. Part of Ellaz, an ad-free family games site in English, Hebrew, Spanish and French.
+  added: 2026-09-02
+  updated: 2026-09-02
+
+- name: Ellaz Match Three
+  originals:
+  - Bejeweled
+  type: similar
```
cxong, line 1001:
```suggestion
  type: clone
```

games/e.yaml
```diff
@@ -927,6 +927,186 @@
...
+  - MIT
+  ai: true
+  info: Minesweeper for phones and desktops, tap to reveal and long-press to flag, with a first tap that is always safe. Part of Ellaz, an ad-free family games site in English, Hebrew, Spanish and French.
+  added: 2026-09-02
+  updated: 2026-09-02
+
+- name: Ellaz Picture Logic
+  originals:
+  - Picross Series
+  type: similar
```
cxong, line 1045:
```suggestion
  type: clone
```

